PR #11, opened 2015-02-09
Adding wheat to the list with SIFT

dbolser-ebi:
I forgot to add this last time... Would like to get it into 'origin' so I don't forget next time (and it should be harmless).

Cheers,
Dan.


willmclaren:
Dan, this is fine, though we have created an eHive pipeline now for VEP dumps, you may find this better anyway: https://github.com/Ensembl/ensembl-variation/blob/release/78/modules/Bio/EnsEMBL/Variation/Pipeline/DumpVEP/DumpVEP_conf.pm

PR #16, opened 2015-04-30
Added genotype phasing when creating objects from adaptor

wspoonr:
This change is needed for genotype phasing to be retrieved correctly.


willmclaren:
Travis failures are unrelated as far as I can tell.

PR #20, opened 2015-05-21
dedupe

dbolser-ebi:
Trying to debug why my transcript variation pipeline is failing and came across this very minor issue.


dbolser-ebi:
Just to clarify, all removed lines are in fact duplicates. See:

    perl -ne 'print unless $x{$_}++' \
        sql/attrib_entries.sql > sql/attrib_entries-dedupe.sql

PR #22, opened 2015-06-26
changes to enable use of eg version in dump vep filename

ckongEbi:


willmclaren:
Thanks. Will need a couple of changes to work with non-EG species, but OK to merge for now.

PR #23, opened 2015-09-18
Add flag to indicate collection eg! species

ckongEbi:


willmclaren:
Looks OK

The flag mechanism is a bit different to others in VEP, i.e. if something defaults to 0 then you should just provide the flag to switch it on i.e. it should be 

--is_multispecies

rather than

--is_multispecies 1

This avoids the confusion suggesting there is a third possible behaviour if you do

--is_multispecies 0

or even

--is_multispecies 24284821222

willmclaren:
(I didn't notice this before as core merged the other PR before I got a chance to look at it!!!)

PR #24, opened 2015-11-05
add eval for bacteria circular region issue

ckongEbi:


willmclaren:
Travis failure is an ongoing issue, nothing to do with this.

No problem with the PR.

PR #27, opened 2016-02-09
Add an install command to the C_code makefile to put the executable in a …

lairdm:
…standardish place for automatic discovery

### DIFF
--- a/C_code/Makefile
+++ b/C_code/Makefile
@@ -2,6 +2,10 @@ CC = gcc
 
 all: calc_genotypes
 
+install: calc_genotypes
+	mkdir -p ../bin
+	cp calc_genotypes ../bin/
+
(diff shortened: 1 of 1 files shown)


willmclaren:
Can you say a little about why this is necessary?

lairdm:
I'm trying to make an easy to use method for users to install the api and populate their paths with the needed libraries and executables. Part of it is a script which hunts for all the possible paths to include, all the "usual" locations.

https://github.com/Ensembl/ensembl/blob/experimental/travis_reorg/activate

So that once they install the repos (and I'm working on a bootstraping script for that too), they simple go:

source ensembl/activate

and PERL5LIB and PATH are populated with all our libraries and executables without too much fiddling. Unfortunately that doesn't work if things aren't in an expected location. Thanks!

PR #30, opened 2016-03-23
making a sensible schema change

dbolser-ebi:


willmclaren:
@ens-lgil is there a good reason for this not being auto-incrementing, something to do with how it syncs with the production DB?

ens-lgil:
I think this is fine to use the auto-increment for the attrib table.
However we don't add data directly into this table anymore: we sync the Variation attrib table with the one from the production DB.

willmclaren:
Perhaps there's some problem with the way it gets populated from Config.pm? I'm just trying to understand why it wasn't set to auto-inc in the first place

ens-lgil:
I think this was because of the design of the native attrib schema: we wanted to keep the IDs stable between releases.

dbolser-ebi:
With regard to keeping the IDs stable between releases, this is what you see happening in the ensembl-variation/sql/attrib_entries.sql file. It's possible that down the line I could use an ID (i.e. for a short_name_attrib_id) that later appears in that file. However, since I shouldn't be re-using that file (i.e. it only ever gets used once at create time), it shouldn't be a problem.

ens-lgil:
Pull request accepted but I needed to change the patch number as the patch "c" was already used.
See commit: https://github.com/Ensembl/ensembl-variation/commit/42b93dfd6a255d3d9b82d0634c9b091306138526

PR #31, opened 2016-04-12
Added moltype description (from what I can guess)

dbolser-ebi:
Please make this clearer if you can :-)

Also, what does "The corresponding variation ID of this variation is different from the one stored in the column variation_id." mean? I think it's trying to say something important, but it isn't clear.

### DIFF
--- a/sql/table.sql
+++ b/sql/table.sql
@@ -227,11 +227,11 @@ create table variation_feature(
 @desc This table allows for a variation to have multiple IDs, generally given by multiple sources.
 
 @column variation_synonym_id	Primary key, internal identifier.
-@column variation_id					Foreign key references to the variation table.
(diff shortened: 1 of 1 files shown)


ens-lgil:
Hi Dan,
I think we should remove the sentence ""The corresponding variation ID of this variation is different from the one stored in the column variation_id." as it is more confusing than helpful.
Concerning the column "moltype", I can't recall why we have this, but it was there when I started. Maybe we can get rid of this column...

willmclaren:
Yeah I don't know what moltype is for either. I suspect this was a Daniel Rios special.

If there's no reference to it anywhere in the API, let's look at removing it for 85.

ens-lgil:
This column is going to be removed foe Ensembl release 85, so there is no need to update the description of this column anymore

PR #32, opened 2016-04-25
Enable VEP cache checking script to be used by Ensembl Genomes

james-monkeyshines:
VEP caches for non-Ensembl sources (i.e. Ensembl Genomes and VectorBase) will be labelled with versions other than the Ensembl version. The latter is still required for things like setting up registries, so the obvious solution is to introduce a new parameter; it's got the same name as the analagous parameter in the VEP script, --cache_version. If not set, it defaults to --version, so this change should have no impact for Ensembl checks.

While adding this functionality, I found what I think is a bug, where variation features were sought even if there is no variation db. So I wrapped that code in a conditional in another commit.

### DIFF
--- a/scripts/misc/healthcheck_vep_caches.pl
+++ b/scripts/misc/healthcheck_vep_caches.pl
@@ -51,6 +51,7 @@ GetOptions(
   
   'dir|d=s',                 
   'version|v=i',
+  'cache_version|c=i',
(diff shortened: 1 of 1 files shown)


willmclaren:
Looks fine to me.

sarahhunt:
Looks fine.

PR #36, opened 2016-09-02
Report plugins errors.

joaoe:
The warn() function call does not output anything. It just supresses errors, in case there are errors in plugins. warning_msg() is the right function to call.


willmclaren:
Thanks @joaoe , looks like we missed these when switching to using warning_msg()

PR #38, opened 2016-11-03
modified fix input vf constrain

Imoteph:
check first if the cache has the same chr naming, before we have actually try to fix them!

As I had the issue that it was complaining not to find any slice as the chr names in only one direction where changed (vf object).

### DIFF
--- a/modules/Bio/EnsEMBL/Variation/Utils/VEP.pm
+++ b/modules/Bio/EnsEMBL/Variation/Utils/VEP.pm
@@ -3218,6 +3218,7 @@ sub validate_vf {
 
     # user specified chr skip list
     return 0 if defined($config->{chr}) && !$config->{chr}->{$vf->{chr}};
+    my $valid_chrs = get_cache_chromosomes($config);
     # fix inputs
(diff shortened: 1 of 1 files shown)


willmclaren:
`$valid_chrs` is not defined yet when you refer to it (it is defined within the `if(defined($config->{cache})) {` block)

This is leading to a failure of the vep.t test in the TravisCI test suite.

Imoteph:
THX sry forgott one line
before
my $valid_chrs = get_cache_chromosomes($config);

willmclaren:
I think this needs a little rearrangement to get the logic in the correct place.

If you don't mind I'll close this PR once I've made a modification that works and fits the current style.

Imoteph:
That´ s fine with, this was just a quick fix for my purpose now. I know that logic here is a bit odd, but I needed a quick solution. Thx

willmclaren:
https://github.com/Ensembl/ensembl-variation/commit/c8d17fb635d36ebc659e3b874b7605aee5d8e8f2 fixes it

PR #40, opened 2016-12-16
fix for obsolete perl syntax (perl 5.24)

balwierz:
Use of keys $hash_ref has been introduced in perl 5.14 and has always been experimental. In 5.24 this syntax throws an error because the support of automatic dereferencing is now removed.

In order to be compatible with perl 5.24 at least one line needs to be modified.


willmclaren:
Not a deliberate use of this syntax I'm sure :-)

Looks fine for me, just need one more to check this and it can be merged.

ens-lgil:
Looks fine for me too.

PR #42, opened 2017-01-13
updating lsf options which where out of date

gnaamati:
Previous LSF options where not consistent with current LSF requests, also updated to rh7


sarahhunt:
Thanks @gnaamati , but some of those numbers don't look correct.

gnaamati:
Thanks @sarahhunt , I have just noticed the case where the -M and [mem=] are not the same which I will fix. 
For example: **-q production-rh7 -M  2000 -R "rusage[mem=15000]"**.
When running this pipeline with the correct numbers it ran OK (I updated them in the command line originally). 
Is there anything else that I might have missed?

willmclaren:
Closing this as I've updated the config now we're running SIFT at EBI too.

https://github.com/Ensembl/ensembl-variation/commit/35ff6464182d8cc1e6fcad47589b715ed091efd2

Of course please let us know if you run into any issues.

PR #46, opened 2017-03-06
missing use for make_path which is used by EG-specific code

danstaines:
Trying to run the existing EG FTP pipeline which includes VEP dumping and found this problem which I imagine would only be found if you supply the flag to tell the dump code that its EG.


willmclaren:
Looks fine, assuming you've tested as we don't run EG code (and this pipeline has been superseded by the one in ensembl-vep). Do you need this on release/88 or release/87 too?

thomasmaurel:
I think this can be closed as the VEP code is now in ensembl-vep and I've removed the EG specific code: https://github.com/Ensembl/ensembl-vep/commit/c4d0b28984d07b9814ec0cf27bc0fece29243bf6

PR #47, opened 2017-03-20
ENSVAR-591 Fix to properly recognize sample synonyms from a VCF back-end

MatBarba:


sarahhunt:
@MatBarba - thanks for the fix, but it is slow for us - we routinely have up to 2700 samples. Using fetch_all_by_name_list for the bulk extraction then individually looking up missing samples would be work better.

MatBarba:
Thanks for the answer! So to solve the bug I reported, I "just" need to recreate the remap the returned samples from fetch_all_by_name_list based on the list of synonyms.
Note however that this method will still be slow if it is given a list of synonyms, because it would then retrieve all samples one by one (=like I did). Most of my VCFs use synonyms, so that may be a problem in the long term (e.g. we have vcf files with 765 samples).

PR #49, opened 2017-04-10
Refactoring LD code

dzerbino:
Hello,

I have had a look at the LD code and found some pretty major time wasting. It would appear that an acceleration of ~10 can be achieved on a simple test, also reducing complexity from O(x^2) to O(x) when searching for the neighbourhood of a SNP. 

I tested it for regression on a simple test, but this could be tested further. 

Cheers,

Daniel

### DIFF
--- a/C_code/Makefile
+++ b/C_code/Makefile
@@ -16,6 +16,7 @@ calc_genotypes: calc_genotypes.o
 calc_genotypes.o:
 	$(CC) -c calc_genotypes.c
 ld_vcf:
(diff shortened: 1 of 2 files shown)


willmclaren:
I've run some tests on this, I'm getting identical results so looks good from my point of view. I'm seeing about a x10 speedup on BCF processing time.

This speedup is massively reduced when using VCF files, but that reflects more the difference in performance between using VCF vs BCF - an argument for switching to BCF on our backend in future when we have ensembl-io support.

@at7 can you review / test as well before we merge please?

at7:
I've run several test cases which produced the same results for both versions and with Daniel's updates mostly faster.

PR #50, opened 2017-04-12
Fix vcf sample synonym

MatBarba:
Reworked version of this fix for ENSVAR-591.
Those 2 commits change 2 things:
1) The way the sample_adaptor get the list of samples from the db: with one request for both samples and sample_synonyms, when before the synonyms were added after the main samples, in a slow loop. This speeds up things significantly when a lot of sample synonyms are used.
2) The sample data can now correctly be retrieved from a vcf. A small bug in the code was hiding the fact that the initial sample_synonym name was forgotten, which resulted in the samples never being mapped (so only new samples unrelated to the db were retrieved).

### DIFF
--- a/modules/Bio/EnsEMBL/Variation/DBSQL/SampleAdaptor.pm
+++ b/modules/Bio/EnsEMBL/Variation/DBSQL/SampleAdaptor.pm
@@ -354,14 +354,7 @@ sub fetch_all_by_name_list {
   
   my $id_str = (@$list > 1)  ? " IN (". join(',', map {'"'.$_.'"'} @$list).")" : ' = \''.$list->[0].'\'';
   
-  my $result = $self->generic_fetch("s.name" . $id_str);
-
-  if (scalar @$result == 0) {
-    foreach my $name (@$list) {
-      push @$result, @{$self->fetch_by_synonym($name)};
-    }
-  }
-  return $result;
+  return $self->generic_fetch("s.name" . $id_str . " OR ss.name" . $id_str);
 }
 
 =head2 fetch_all_by_Population
--- a/modules/Bio/EnsEMBL/Variation/VCFCollection.pm
+++ b/modules/Bio/EnsEMBL/Variation/VCFCollection.pm
@@ -492,13 +492,10 @@ sub get_all_Samples {
     # some may not be in DB
     foreach my $vcf_sample_name (@$vcf_sample_names) {
       # Use the main sample name to retrieve its metadata
-      my $sample_name = $vcf_sample_name;
-      if (exists $synonyms{ $vcf_sample_name }) {
-        $sample_name = $synonyms{ $vcf_sample_name };
-      }
+      my $sample_name = $synonyms{ $vcf_sample_name } || $vcf_sample_name;
       
       # either use the DB one or create one
-      my $sample = $sample_objs{ $prefix.$sample_name } //
+      my $sample = $sample_objs{ $prefix.$sample_name } ||
         Bio::EnsEMBL::Variation::Sample->new_fast({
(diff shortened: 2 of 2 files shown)


modules/Bio/EnsEMBL/Variation/VCFCollection.pm
```diff
@@ -474,16 +474,31 @@ sub get_all_Samples {
...
+    }
+
     # some may not be in DB
-    foreach my $sample_name(@$sample_names) {
+    foreach my $vcf_sample_name (@$vcf_sample_names) {
+      # Use the main sample name to retrieve its metadata
+      my $sample_name = $vcf_sample_name;
+      if (exists $synonyms{ $vcf_sample_name }) {
+        $sample_name = $synonyms{ $vcf_sample_name };
+      }
```
willmclaren: This would be neater written as:

`my $sample_name = $synonyms{$vcf_sample_name} || $vcf_sample_name;`

modules/Bio/EnsEMBL/Variation/VCFCollection.pm
```diff
@@ -474,16 +474,31 @@ sub get_all_Samples {
...
+   
+    # Get the list of sample synonyms
+    my %synonyms;
+    for my $sample_name (keys %sample_objs) {
+      my $sample = $sample_objs{$sample_name};
+      for my $syn ($sample->get_all_synonyms) {
+        $synonyms{$syn->[0]} = $sample->name;
+      }
+    }
+
```
willmclaren: There might be a shorter way to write this with a map, but it's fine as is.

Assuming this is still fast as synonyms are loaded at the time the sample object is created?

modules/Bio/EnsEMBL/Variation/VCFCollection.pm
```diff
@@ -474,16 +474,31 @@ sub get_all_Samples {
...
+    foreach my $vcf_sample_name (@$vcf_sample_names) {
+      # Use the main sample name to retrieve its metadata
+      my $sample_name = $vcf_sample_name;
+      if (exists $synonyms{ $vcf_sample_name }) {
+        $sample_name = $synonyms{ $vcf_sample_name };
+      }
+      
       # either use the DB one or create one
-      my $sample = $sample_objs{$prefix.$sample_name} ||
+      my $sample = $sample_objs{ $prefix.$sample_name } //
```
willmclaren: No real need to change from `||` to `//`?

I prefer to avoid it if possible as the operator is not supported in old Perl versions.

modules/Bio/EnsEMBL/Variation/DBSQL/SampleAdaptor.pm
```diff
@@ -354,14 +354,7 @@ sub fetch_all_by_name_list {
...
   
-  my $result = $self->generic_fetch("s.name" . $id_str);
-
-  if (scalar @$result == 0) {
-    foreach my $name (@$list) {
-      push @$result, @{$self->fetch_by_synonym($name)};
-    }
-  }
-  return $result;
+  return $self->generic_fetch("s.name" . $id_str . " OR ss.name" . $id_str);
```
willmclaren: This makes much more sense, and allows for the case where the number fetched was >0 but <`@list`.

willmclaren:
The change to SampleAdaptor.pm is now much more sensible.

The Travis fail is because of the change to htslib pre-requisites.

MatBarba:
I've added a small commits with the suggested changes.
I believe get_all_synonyms() is fast enough since it doesn't make any call to the db, i.e. I got 7ms for 765 samples, compared to the 212 ms spent to retrieve the samples from the DB with fetch_all_by_name_list.

PR #54, opened 2017-05-03
Bug when assessing whether variant of interest was found in LD code

dzerbino:
Hello @at7, 

here is the bug fix you requested, an '=' character was missing. 

Thanks for the good test case, it helped tremendously,

Daniel


at7:
All test cases pass again. Thank you

PR #57, opened 2017-07-13
Registry and divisions support

thomasmaurel:
The pull request contains the following major changes:

The Ensembl Production Species Factory is now used to retrieve the variation databases from a registry file. 
Updated default location for so_file to use the same file that Production use to generate the ontology database.
gvf_validator, vcf_validator and vcf_sort are now part of the ensembl environment. 
Restructured the pipeline so that each module can run on one species only. 
Deprecated the data_dumps_config.json file and the SpeciesFactory.pm, this data is now passed through the pipeline as a param. 
Removed Report.pm as it's not used by any pipeline. 
The species division is now stored into the config param, all the modules have been updated to check it's value, if it's empty pipeline_dir will be used. If we have a value for a division, the data will be stored into pipeline_dir and then into a division directory. 
Updated checks for vcf, gvf validator to properly report the error if missing, added a check for vcf-sort in PreRunChecks.pm. 
Minor tidying up for many modules. 
Updated dataflow_output_id to work with new pipeline flows.
Updated README file.

I did a test run on ensembl 89 and EG metazoa 36. Output files are the same as the live version (excluding data recently removed by Anja in https://github.com/Ensembl/ensembl-variation/commit/3ec5c9b0fb143baba4da28e3ff13a9bc3df4f957
Please see: 
/hps/nobackup/production/ensembl/maurel/variation_ensembl_test_dumps_89
/hps/nobackup/production/ensembl/maurel/variation_metazoa_test_dumps_36

I will update the confluence documentation as soon as the pull request is merged.


at7:
Thomas explained all code changes to me.

PR #59, opened 2017-07-22
Plugged a memory leak in the results cache of BaseVepTabixPlugin

mkschuster:
Hi Will,

The linked commit plugs a memory leak in the BaseVepTabixPlugin module that showed with the CADD VEP plugin. For a genome-wide VCF (around 16M variants), VEP previously required in excess of 105 GiB, which is a bit much. With this fix applied, it only requires around 750 MiB throughout.

Thanks,
Michael


willmclaren:
Thanks Michael - pretty daft copy/paste error there!

PR #62, opened 2017-08-25
Removed options from parent package

thomasmaurel:
Minor changes to the GVF/VCF pipeline. One bug fix for EG divisions and minor clean up of the config file. I've tested these changes on Plants EG 37.


at7:
ok

PR #66, opened 2017-09-27
initial synonym importer - loads PharmGKB data

sarahhunt:
Requires new API method to store synonyms for existing variant

### DIFF
--- a/modules/Bio/EnsEMBL/Variation/DBSQL/VariationAdaptor.pm
+++ b/modules/Bio/EnsEMBL/Variation/DBSQL/VariationAdaptor.pm
@@ -191,8 +191,10 @@ sub store_synonyms{
       my $source_adaptor = $self->db->get_SourceAdaptor();
       my $source = $source_adaptor->fetch_by_name($source_name);
 
+      throw("No source found for name $source_name") unless defined $source;
+
       foreach my $name (keys %{$var->{synonyms}->{$source_name}}){
-        $sth->execute($var->{dbID}, $name, $source->dbID()) ;
+        $sth->execute($var->dbID(), $name, $source->dbID()) ;
       }
     }
     return $var;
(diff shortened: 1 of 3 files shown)


modules/Bio/EnsEMBL/Variation/DBSQL/VariationAdaptor.pm
```diff
@@ -165,6 +165,37 @@ sub store {
...
+     ) VALUES (?,?,?)
+    });
+
+    foreach my $source_name (keys %{$var->{synonyms}} ){  ### update this to allow storage of id
+
+      my $source_adaptor = $self->db->get_SourceAdaptor();
+      my $source = $source_adaptor->fetch_by_name($source_name);
+
+      foreach my $name (keys %{$var->{synonyms}->{$source_name}}){
+        $sth->execute($var->{dbID}, $name, $source->dbID()) ;
```
willmclaren: If source name not in the DB it will die here, should check source is defined and throw after line 192?

Also any reason to use `$var->{dbID}` and `$source->dbID()`? Probably best to make both method calls?

scripts/import/import_variant_synonyms
```diff
@@ -0,0 +1,214 @@
...
+        -data_types  => ['variation_synonym']
+      );
+    eval{$source_adaptor->store($source);} ; die "ERROR storing source: $@\n" unless $@ eq ''; 
+  }
+  return $source;
+}
+
+sub usage{
+
+  die "\nUsage : import_variant_synonyms -registry [registry file] -source_name [name]
```
willmclaren: Other import scripts are named [script].pl, we should probably be consistent

scripts/import/import_variant_synonyms
```diff
@@ -0,0 +1,214 @@
...
+\n\tOptions: -clean remove of data
```
willmclaren: "-clean remove of data" ?

PR #67, opened 2017-09-27
Add overridden caching versions of seq_region_start, seq_region_end

willmclaren:
Required to fix VEP issue where seq_region_start gets looked up from
DB if VF is on a circular slice.

transfer and transform methods are overridden to prevent issues
with caching.


willmclaren:
See also https://github.com/Ensembl/ensembl-vep/pull/83

PR #70, opened 2017-10-03
Add option for HGVS parser to replace given ref with seq looked up from DB

willmclaren:


modules/Bio/EnsEMBL/Variation/DBSQL/VariationFeatureAdaptor.pm
```diff
@@ -1927,7 +1949,10 @@ sub _hgvs_from_components {
  
   else {
     # If the reference from the sequence does not correspond to the reference given in the HGVS notation, throw an exception
```
sarahhunt: It would be clearer to move this comment a couple of lines down now.

modules/Bio/EnsEMBL/Variation/DBSQL/VariationFeatureAdaptor.pm
```diff
@@ -1927,7 +1949,10 @@ sub _hgvs_from_components {
...
   else {
     # If the reference from the sequence does not correspond to the reference given in the HGVS notation, throw an exception 
-    if (defined($ref_allele) && $ref_allele ne $refseq_allele){        
+    if($replace_ref && defined($ref_allele)) {
+      $ref_allele = $refseq_allele;
+    }
+    elsif (defined($ref_allele) && $ref_allele ne $refseq_allele){        
       throw ("Reference allele extracted from $reference:$start-$end ($refseq_allele) does not match reference allele given by HGVS notation $hgvs ($ref_allele)");
     }
     if($DEBUG==1){print "Reference allele: $refseq_allele expected allele: $ref_allele\n";}
```
sarahhunt: The debug would be more useful before the ref allele is reset

PR #74, opened 2017-10-11
Feature/sift species factory 90

dbolser-ebi:
Do you have any tests for this ;-)


willmclaren:
Can you please do this PR against master instead?

We don't have any testing framework for eHive pipelines in place, so we're relying on you having thoroughly tested the code yourselves.

PR #82, opened 2017-11-01
At7/ld tool api support

at7:
Update API LD calculation functionality. We need to be able to compute pairwise LD values for more than 2 variation features. We need to be able to filter results by r2 and d_prime values. Add more tests.

### DIFF
--- a/C_code/ld_vcf.c
+++ b/C_code/ld_vcf.c
@@ -501,12 +501,11 @@ int main(int argc, char *argv[]) {
 
   // parse args
   int c;
-  char *files[2];
-  char *regions[2];
+  char *filestr = NULL;
+  char *regionstr = NULL;
   char *samples_list = NULL;
-  char *variants_file;
+  char *variants_file = NULL;
   char *variant = NULL;
-  int numfiles = 0;
   int numregions = 0;
   int windowsize = WINDOW_SIZE;
 
@@ -514,8 +513,7 @@ int main(int argc, char *argv[]) {
     static struct option long_options[] = {
       {"file",    required_argument, 0, 'f'},
       {"region",  required_argument, 0, 'r'},
-      {"file2",   required_argument, 0, 'g'},
-      {"region2", required_argument, 0, 's'},
+      {"numregions", required_argument, 0, 's'},
       {"samples", required_argument, 0, 'l'},
       {"window",  required_argument, 0, 'w'},
       {"variant", required_argument, 0, 'v'},
@@ -526,7 +524,7 @@ int main(int argc, char *argv[]) {
     /* getopt_long stores the option index here. */
     int option_index = 0;
 
-    c = getopt_long (argc, argv, "f:g:l:r:s:w:v:n:", long_options, &option_index);
+    c = getopt_long (argc, argv, "f:l:r:s:w:v:n:", long_options, &option_index);
 
     /* Detect the end of the options. */
     if (c == -1)
@@ -535,19 +533,15 @@ int main(int argc, char *argv[]) {
     switch (c) {
 
       case 'f':
-        files[numfiles++] = optarg;
-        break;
-
-      case 'g':
-        files[numfiles++] = optarg;
+        filestr = optarg;
         break;
 
       case 'r':
-        regions[numregions++] = optarg;
+        regionstr = optarg;
         break;
 
       case 's':
-        regions[numregions++] = optarg;
+        numregions = (int) atoi(optarg);
         break;
 
       case 'l':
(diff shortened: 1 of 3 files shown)


modules/Bio/EnsEMBL/Variation/DBSQL/LDFeatureContainerAdaptor.pm
```diff
@@ -102,14 +102,42 @@ sub max_snp_distance {
   return $self->{'max_snp_distance'};
 }
 
-sub executable {
-  my $self = shift;
-  $BINARY_FILE = shift if @_;
-  unless( $BINARY_FILE ) {
-    my $binary_name = 'calc_genotypes';
-    ($BINARY_FILE) = grep {-e $_} map {"$_/$binary_name"} split /:/,$ENV{'PATH'};
+=head2 r2
```
willmclaren: It was not immediately obvious to me what this meant - I think the method (and const) should be called minimum_r2 or min_r2 to make it clear. Same for dprime.

modules/Bio/EnsEMBL/Variation/DBSQL/LDFeatureContainerAdaptor.pm
```diff
@@ -454,6 +435,8 @@ sub _fetch_by_Slice_VCF {
           $sample_count
         ) = split /\s/;
 
+        # filter by r2 and d_prime values
+        next if ($r2 < $self->r2 || $d_prime < $self->d_prime);
```
willmclaren: You could put $self->r2 and $self->d_prime in variables outside the loop to avoid calling the methods every loop.

C_code/ld_vcf.c
```diff
@@ -574,25 +568,44 @@ int main(int argc, char *argv[]) {
         abort ();
     }
   }
+  char *files[numregions];
```
willmclaren: I don't know how off the top of my head but there must be a neater way to do this without manually specifying num_regions as a command line option.

You can see there's code elsewhere in this file for dynamically expanding the memory used by an array.

Having said that, it works, and I don't expect many people to be using this executable outside our API calls.

modules/t/LDFeatureContainerAdaptor.t
```diff
@@ -129,21 +129,87 @@ foreach my $ld_value (@$ld_values) {
...
+throws_ok { $ldfca->fetch_by_Slice([$vf1]); } qr/Bio::EnsEMBL::Slice arg expected/, 'fetch_all_by_Slice Throw on wrong argument';
+
+throws_ok { $ldfca->fetch_all_by_Variation('Variation'); } qr/Bio::EnsEMBL::Variation::Variation arg expected/, 'fetch_all_by_Variation Throw on wrong argument';
+throws_ok { $ldfca->fetch_by_VariationFeature('VariationFeature'); } qr/Bio::EnsEMBL::Variation::VariationFeature arg expected/, 'fetch_by_VariationFeature Throw on wrong argument';
+throws_ok { $ldfca->fetch_by_VariationFeatures(['VariationFeature']); } qr/Bio::EnsEMBL::Variation::VariationFeature arg expected/, 'fetch_by_VariationFeatures Throw on wrong argument';
+throws_ok { $ldfca->fetch_by_VariationFeatures('VariationFeature'); } qr/Listref of Bio::EnsEMBL::Variation::VariationFeature args expected/, 'fetch_by_VariationFeatures Throw on wrong argument';
 
-done_testing();
 
+
```
willmclaren: 6 newlines? /petty

modules/t/LDFeatureContainerAdaptor.t
```diff
@@ -129,21 +129,87 @@ foreach my $ld_value (@$ld_values) {
...
   my $d_prime = $ld_value->{d_prime};
-  cmp_ok($result->{$variation_name2}->{r2}, '==', $r2, " fetch_by_VariationFeature r2 for $variation_name2 and rs1333049");
-  cmp_ok($result->{$variation_name2}->{d_prime}, '==', $d_prime, " fetch_by_VariationFeature d_prime for $variation_name2 and rs1333049");
+  cmp_ok($result->{$variation_name2}->{r2}, '==', $r2, "fetch_by_VariationFeature r2 for $variation_name2 and rs1333049");
+  cmp_ok($result->{$variation_name2}->{d_prime}, '==', $d_prime, "fetch_by_VariationFeature d_prime for $variation_name2 and rs1333049");
 }
 
-#fetch_by_VariationFeatures
-#{ population_id => 102179, vf1_name => 'rs1333047', vf2_name => 'rs72655407', r2 => 0.063754, d_prime => 0.999996 },
+#fetch_by_VariationFeature, population is not defined
```
willmclaren: Good job adding all these extra tests - this module was hideously undertested before.

C_code/ld_vcf.c
at7: I couldn't get dynamic arrays to work. I'll get back to it. For now I added a default value for number of allowed regions. In case someone should run the script outside of the module. I got an 'argument list too long' error when allowing 1000 regions. I set the value to 100 which works. If it becomes a common use case to pass a long list of variant IDs I'll spend more time on implementing a better solution.

willmclaren: That seems reasonable

PR #90, opened 2017-11-30
Skip the AnimalQTL entries with the map type 'linkage' or the start c…

ens-lgil:
…oordinate equal to 0


ens-lgil:
This is for e92 (no need to do mpush to release/91)

PR #93, opened 2017-12-12
MySQL version review

helensch:
Modifications for differences in MySQL versions

- handling of GROUP BY different between 5.6 and 5.7. 
  Addresses issues in JIRA ENSVAR-881

- storage_engine deprecated 5.7.5


willmclaren:
Looks fine. Do you know if there is any performance difference between `DISTINCT ... ORDER BY` vs `GROUP BY`?

helensch:
The MySQL explain excution plan is the same (or very similar) for a DISTINCT ... ORDER BY and a GROUP BY for the changes made. 

The ORDER BY on DISTINCT adds a "Using filesort". This is also in the GROUP BY.

PR #94, opened 2017-12-13
store HGVS reference string for access in VEP

sarahhunt:


modules/Bio/EnsEMBL/Variation/TranscriptVariationAllele.pm
```diff
@@ -729,12 +729,32 @@ sub hgvs_transcript {
...
+
+=cut
+
+sub hgvs_transcript_reference{
+
+  my $self = shift;
+  defined $self->{hgvs_t_ref} ? return $self->{hgvs_t_ref} : undef;
+
+}
+
```
willmclaren: Could this method call `$self->hgvs_transcript` if `$self->{hgvs_t_ref}` is not defined, rather than returning `undef`? I think that would be neater, even if that avenue doesn't end up getting used.

sarahhunt: Yes - I was not sure how often hgvs_transcript would return nothing for a TVA and how slow it would be to call it twice. Do you reckon it's significant?

willmclaren: You could set `$self->{hgvs_t_ref}` to undef at the beginning of `hgvs_transcript()`, then check `exists $self->{hgvs_t_ref}` rather than `defined`

PR #98, opened 2018-01-02
Added region length check

aparton:
Exception is thrown where HGVS notation spanning more than 5kb is requested


sarahhunt:
A test would be nice.

sarahhunt:
This looks fine, but before it goes in master, can you ensure VEP/Parser/HGVS.pm handles the exception graciously. If one variant in the middle of a set is a typo, the correct ones after it should be analysed.

PR #100, opened 2018-01-03
Attrib update

helensch:
Update attrib_entries.sql to match what in current variation databases. See JIRA ENSVAR-821.


ens-lgil:
Actually we don't use this attrib_entries.sql script anymore as we use the tables from the production database (attrib, attrib_type and attrib_set), unless we use this script to populate the attrib tables as a starting point when we create a new variation DB

helensch:
The script was updated for use when creating new variation DB as attribs e.g. TOPMED and GNOMAD were needed.

ens-lgil:
At some point in the future we should get the "attrib" data directly from the Production DB instead of using this SQL file: this will prevent the maintenance of an extra SQL file.

PR #101, opened 2018-01-09
Rewrite some of the API (and test) for the StructuralVariationPopulat…

ens-lgil:
…ionFrequency object, in order to calculate correctly the human frequencies on the chromosomes X and Y


ens-lgil:
Right! I can "merge" most of the code from the method "allele_count" into "allele_count_by_class_SO_term".
Will try to do it in a next PR

PR #104, opened 2018-01-12
Change the zygosity data type from a string (e.g. homozygous) to an i…

ens-lgil:
…nteger (e.g. 2)


modules/t/test-genome-DBs/homo_sapiens/variation/table.sql
```diff
@@ -168,7 +168,7 @@ CREATE TABLE `meta` (
   PRIMARY KEY (`meta_id`),
   UNIQUE KEY `species_key_value_idx` (`species_id`,`meta_key`,`meta_value`),
   KEY `species_value_idx` (`species_id`,`meta_value`)
-) ENGINE=MyISAM AUTO_INCREMENT=99 DEFAULT CHARSET=latin1;
```
at7: I wonder if removing the charset is causing a problem in travis. We will see.

PR #107, opened 2018-01-15
Disable addition of a had_no_warnings test via done_testing

at7:
The tests don't fail when I run them on the farm. I added the change which seems to let the tests pass. Not entirely sure what is going on. Travis is probably throwing a warning message which is causing problems.

### DIFF
--- a/modules/t/LDFeatureContainerAdaptor.t
+++ b/modules/t/LDFeatureContainerAdaptor.t
@@ -17,7 +17,7 @@ use strict;
 use warnings;
 use Test::More;
 use Test::Exception;
-use Test::Warnings qw(warning);
+use Test::Warnings ':no_end_test';
 use Bio::EnsEMBL::Test::TestUtils;
 use Bio::EnsEMBL::Test::MultiTestDB;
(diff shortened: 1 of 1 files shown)


modules/t/LDFeatureContainerAdaptor.t
```diff
@@ -17,7 +17,7 @@ use strict;
 use warnings;
 use Test::More;
 use Test::Exception;
-use Test::Warnings qw(warning);
+use Test::Warnings ':no_end_test';
```
ens-lgil: Should we simply remove the line instead ?

at7: I'm actually testing that a warning is returned from the adaptor in certain cases.

at7:
The warning in travis that is causing the test to fail is 

> For travis_test_db_homo_sapiens_variation_20180115_143048@127.0.0.1 there is a difference in the software release (92) and the database release (91)

Not sure where the information about the software release is stored or retrieved from.

ens-lgil:
It looks like a Core warning message.

PR #113, opened 2018-01-23
Master parseline

aparton:
Moved parse_line method from old VEP.pm into new file


ens-lgil:
Should it go to release/92 as well ?

sarahhunt:
There is a lot of stuff that isn't needed here, for example it is exporting functions it doesn't contain. We only need the VCF parsing - not pileup or format detection, etc. This PR causes a 1% drop in coverage which I assume is mostly due to the redundant lines, though we may need test updates.

willmclaren:
It seems to me this is only half the job - really there should be one method that transforms a VCF entry into a VariationFeature (or StructuralVariationFeature). At the moment this is done both here and in the VEP::Parser::VCF method.

The issue is that different use cases may require slightly different outcomes. Here's some thoughts off the top of my head:

* A VariationFeature is a subclass of the generic Feature class, which inherently requires it to be attached to a slice. However, for VEP's purposes we don't want this as we may want to lazy-fetch the slice later.

* There's also the issue of dealing with genotypes. At the moment this parse_vcf method is not great as it (given the correct config) will return one VariationFeature object per individual, as opposed to one VariationFeature with multiple genotypes attached (which it probably should). To change this will require some change to the VEP code that deals with this.

* How to deal with allele normalisation

* might the method want to only create a VF for one from multiple ALTs?

I guess you're beginning to see why I didn't just do this quickly...

PR #116, opened 2018-02-06
Script to check the VARIATION, STRUCTURAL_VARIATION and PHENOTYPE exa…

ens-lgil:
…mples from the ini-files of the public_plugins repo


sarahhunt:
This is very useful!

As the same ini files are used for GRCh37 and GRCh38, it would be good to be able to check human GRCh37 too rather than skip it.

ens-lgil:
Just removed the skip for the GRCh37 DB

PR #117, opened 2018-02-06
Fix the VEP.pm module by removing/updating the calls to deprecated Re…

ens-lgil:
…gulation methods


ens-lgil:
Same issue we have in ensembl-vep (cf https://github.com/Ensembl/ensembl-vep/pull/140)

PR #118, opened 2018-02-08
Methods to retrieve HGVS exon coordinates and intron distances

stekaz:
Hi, I'd like to be able to filter some VEP output using the HGVS intron offset values as a measure of intronic distance. I couldn't see an easy way to retrieve these and the exon coordinates without having to either parse the HGVS string or re-calculate them. I added some methods for these, but if there's a better way to do this please let me know.

### DIFF
--- a/modules/Bio/EnsEMBL/Variation/TranscriptVariationAllele.pm
+++ b/modules/Bio/EnsEMBL/Variation/TranscriptVariationAllele.pm
@@ -1,7 +1,7 @@
 =head1 LICENSE
 
 Copyright [1999-2015] Wellcome Trust Sanger Institute and the EMBL-European Bioinformatics Institute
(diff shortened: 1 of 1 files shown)


aparton:
Hi,

Thank you for this. Currently, there is no easy way to retrieve the information that you require, so we are happy to add your methods. To be added to the system according to our procedures, the PR will need to be on the master branch, and tests will need to be added to our testing suite. I'm happy to do this for you, but it will take a couple of days and will require the PR to be on the master branch.

I'll let you know when these changes have been made. 

Kind Regards,
Andrew

aparton:
Hi Stekaz,

These changes have now been made to the master branch- please see https://github.com/Ensembl/ensembl-variation/pull/124.

Thank you, and kind regards,
Andrew

stekaz:
Thanks Andrew, much appreciated!

PR #120, opened 2018-02-12
patch fix - set as MyISAM

sarahhunt:


ens-lgil:
Hi Sarah,
It should be pushed to postreleasefix/92 instead of master

PR #122, opened 2018-02-12
Citation loader

sarahhunt:
Updates to:
 - store the date the pipeline is run in meta 
 - improve comments and log file names
 - don't check citations which are already loaded
 - use https 
 - only update display statuses which are wrong
 - write out variants which were not found in the same format they were input
 - add goat aliases


scripts/import/import_EPMC.pl
```diff
@@ -277,7 +295,7 @@ sub get_epmc_data{
     return undef unless defined $id && $id =~/\d+/;
 
     my $xs   = XML::Simple->new();
-    my $server = 'http://www.ebi.ac.uk/europepmc/webservices/rest/';
+    my $server = 'https://www.ebi.ac.uk/europepmc/webservices/rest/';
```
ens-lgil: Would it make sense to move this "$server" variable outside the method in order to avoid having to create this variable each time we call the "get_epmc_data" method ?

PR #123, opened 2018-02-13
make use of include_failed_variations flag in VariationFeatureAdaptor…

ima23:
… and fixed typo

### DIFF
--- a/modules/Bio/EnsEMBL/Variation/DBSQL/VariationFeatureAdaptor.pm
+++ b/modules/Bio/EnsEMBL/Variation/DBSQL/VariationFeatureAdaptor.pm
@@ -294,6 +294,8 @@ sub update {
 sub fetch_all {
     my $self = shift;
     my $constraint = 'vf.somatic = 0';
+    # Add the constraint for failed variations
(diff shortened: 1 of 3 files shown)


ens-lgil:
Should it go into the release/92 ?

ima23:
@ens-lgil: thanks for checking. It's not urgent, would be nice 92 but 93 is also ok.

at7:
It would be worth checking why VariationFeatureAdaptor::_default_where_clause is not setting the failed status.

at7:
I wouldn't worry about why _default_where_clause is not used. It is only relevant for the fetch_all methods which are not useful for the amount of data we have anyway. You could call VariationFeatureAdaptor::fetch_all_by_Slice_constraint from VariationFeatureAdaptor::fetch_all and VariationFeatureAdaptor::fetch_all_somatic. But not necessary in my opinion.

at7:
This needs to go to master and release/92

ima23:
Closed for a cleaner PR

ima23:
The final tests are in PR #141. The solution to the discrepancy was to clear the cache or not use any caching. Default mode uses a cache.

PR #133, opened 2018-03-09
add tests for failed flag

ima23:


ima23:
Closed until further inspection.

ima23:
The final tests are in PR https://github.com/Ensembl/ensembl-variation/pull/141. The solution to the discrepancy was to clear the cache or not use any caching. Default mode uses a cache.

PR #136, opened 2018-03-15
add new attribs

ima23:


ens-lgil:
So I believe we will need to patch the schema to add these new attrib_id to the column "mapped_by_attrib" in the table "phenotype_ontology_accession"

ima23:
@ens-lgil good point! added the patch in next PR.

PR #137, opened 2018-03-15
add required patches for new ontology_mapping attribs

ima23:


sarahhunt:
Thanks @ima23 ! Could you update table.sql too?

ens-lgil:
@ima23, you will need to update those lines as well in table.sql:
https://github.com/Ensembl/ensembl-variation/blob/master/sql/table.sql#L1789

PR #139, opened 2018-03-16
import p_values with less-than character

ima23:


scripts/import/import_phenotype_data.pl
```diff
@@ -2955,6 +2955,10 @@ sub convert_p_value {
       $sci_pval = $pval;
     #$sci_pval = sprintf("%.2e",$pval); # e.g. 0.002 => 2,30e-3
     }
+    # If a less than format is found (e.g. <0.05)
+    elsif ($pval =~ /^<\s*(\d+.*)/){
```
ens-lgil: Wondering if it would be simplier (and if this would work!) to merge the conditions `$pval =~ /^\d+/` and `$pval =~ /^<\s*(\d+.*)/`
into `$pval =~ /(\d+.*)/`

ima23: Thanks @ens-lgil! The current examples work with the simplified if/else regex combination. I've made the suggested change.

PR #141, opened 2018-03-16
additional tests for variation fetch by slice, with/without cache

ima23:


ima23:
@ens-lgil yes, the counting issue was due to another test storing an entry in the database. I've added   the needed wrapper methods to the other test in the next commit.

PR #154, opened 2018-04-10
Add new evidence status (TOPMED and gnomAD)

ens-lgil:


at7:
I will merge it into master and release/93

PR #155, opened 2018-04-11
select vf by allele_string

at7:
Finally found the reason for the failed tests for 5.26


at7:
Merged to master and release/93

PR #156, opened 2018-04-17
test with defined

at7:
This will fix a problem with the[ variation table view](https://plants.ensembl.org/Oryza_sativa/Gene/Variation_Gene/Table?db=core;g=Os05g0113900;r=5:738208-739079;t=Os05t0113900-0).


at7:
travis fails because of a regulation error not related to the PR

PR #163, opened 2018-04-23
Check that you've got a Transcript object before calling methods

james-monkeyshines:


james-monkeyshines:
When creating dump files for plants, this bit of code would raise an error for a couple of species, because the $tran variable would be a hash, not an object, but with the 'peptide' key undefined. This would mean that $peptide ended up as 'undef', which the code subsequently assumed to mean that $tran was an object.

james-monkeyshines:
Do you know which species that was? The failures I saw were for Hordeum vulgare (barley). (I've just realise that the failures I saw for the other species (cress) were memory related, and not connected to this issue.)

at7:
It is in Oryza sativa Japonica (http://plants.ensembl.org/Oryza_sativa/Gene/Variation_Gene/Table?db=core;g=Os05g0113900;r=5:738208-739079;t=Os05t0113900-01)

james-monkeyshines:
OK, this isn't the same data issue, the barley stable IDs are consistent. But since you imply this is caused by a data problem, I'll kick it over to the plants team for them to investigate.

PR #168, opened 2018-04-30
Post processing script to cleanup the phenotype related table in all …

ens-lgil:
…the variation databases


at7:
This will make our live much easier. Can you add the option to use a registry file and read connection parameters from file rather than passed as script arguments? Thank you

PR #178, opened 2018-05-31
Update most of the Variation web documentation scripts, mainly becaus…

ens-lgil:
…e of the new structure of the Variation web documentation. Also created a new script to generate it


scripts/misc/generate_variation_set_table.pl
```diff
@@ -22,7 +22,7 @@ =head1 CONTACT
   developers list at <http://lists.ensembl.org/mailman/listinfo/dev>.
 
   Questions may also be sent to the Ensembl help desk at
-  <http://www.ensembl.org/Help/Contact>.
+  <helpdesk.org>.
```
sarahhunt: This doesn't look right; should it not be http://www.ensembl.org/Help/Contact?

ens-lgil: Strange URL indeed! 
FIxed in the next commit.
Thanks!

PR #182, opened 2018-06-11
add count by transcript for TranscriptVariation

sarahhunt:
Attemped fix for https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-4496.


at7:
merged with release/93 and master

PR #185, opened 2018-06-19
Dump pipeline prf93

at7:
Update code to deal with dumping some types of human data dumps per chromosome. Update readme files and provide readme file for human only.


at7:
We can schedule a code review for July and someone else can run the pipeline for release/94. For now it would be good to get the code into release/93 because it was used for generating release/93 data.

PR #190, opened 2018-06-25
Update vfa

at7:


modules/Bio/EnsEMBL/Variation/DBSQL/VariationFeatureAdaptor.pm
```diff
@@ -481,6 +481,39 @@ sub fetch_all_somatic_by_Slice_Source {
   return $self->fetch_all_somatic_by_Slice_constraint($slice, $constraint);
 }
 
+
+=head2 fetch_all_by_Slice_Source
+  Arg [1]    : Bio::EnsEMBL::Slice $slice the slice from which to obtain features
+  Arg [2]    : Bio::EnsEMBL::Variation::Source $source only return somatic mutations for the given source
```
ens-lgil: Might need to fix this: I believe this concerns all the variant sources

PR #192, opened 2018-06-29
SequenceOntologyMapper deprecation update

tgrego:
`Bio::EnsEMBL::Utils::SequenceOntologyMapper` is being deprecated with https://github.com/Ensembl/ensembl/pull/246
That functionality is being replaced with Bio::EnsEMBL::Feature::feature_so_acc, and all features will be able to use that method provided they declare a constant SO_ACC.
I'm making sure all the 'features' that were supported on SequenceOntologyMapper are compliant so that pipelines that use it can be updated seamlessly.
VariationFeature and StructuralVariationFeature are some of the features that require SO_ACC.

### DIFF
--- a/modules/Bio/EnsEMBL/Variation/BaseVariationFeature.pm
+++ b/modules/Bio/EnsEMBL/Variation/BaseVariationFeature.pm
@@ -56,6 +56,7 @@ use warnings;
 package Bio::EnsEMBL::Variation::BaseVariationFeature;
 
 use Bio::EnsEMBL::Feature;
+use Bio::EnsEMBL::Variation::Utils::Constants qw($SO_ACC_MAPPER);
 
 our @ISA = ('Bio::EnsEMBL::Feature');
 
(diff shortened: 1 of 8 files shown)


ens-lgil:
@tgrego,

We would prefer to have the 2 constants stored in our module [Config.pm](https://github.com/Ensembl/ensembl-variation/blob/master/modules/Bio/EnsEMBL/Variation/Utils/Config.pm), for instance something like this:
```
our %SO_ACC_MAPPER =  = (
  'variant' => "SO:0001537",
  'sv'        => "SO:0001060"
);
```
Then we will need to have it defined in our [Constants.pm](https://github.com/Ensembl/ensembl-variation/blob/master/modules/Bio/EnsEMBL/Variation/Utils/Constants.pm) module (we have a script to build it from Config.pm).
Finally we could use it in the different modules you updated:
`use Bio::EnsEMBL::Variation::Utils::Constants qw(%SO_ACC_MAPPER);`

Please let me know if you need more information.
Laurent

tgrego:
I've added the constant definition to Config, and regenerated Constants.
Decided to go for override of core feature_so_acc.
If other features require the method the constant in Config needs to be updated, and they must declare the method (or inherit from these 2).

tgrego:
for release/94 please, it has gone into core already...
Thank you!

tgrego:
One last commit to reduce duplication of code.
single feature_so_acc override on BaseVariationFeature.

PR #196, opened 2018-07-02
implement PhenotypeFeature summary_as_hash method

ima23:


modules/Bio/EnsEMBL/Variation/PhenotypeFeature.pm
```diff
@@ -1188,5 +1188,46 @@ sub date_last_evaluated{
...
+  my $self = shift;
+
+  my $summary_ref = $self->SUPER::summary_as_hash;
+  $summary_ref->{$self->type()} = $self->object_id;
+  $summary_ref->{'description'} = $self->phenotype->description;
+  $summary_ref->{'source'} = $self->source_name;
+  $summary_ref->{'location'} = $self->seq_region_name.":".$self->seq_region_start."-".$self->seq_region_end;
+
+  my %attribs;
+  $attribs{'risk_allele'} = $self->risk_allele                            if defined $self->risk_allele;
```
ens-lgil: Wondering if it would be easier to simply use `$self->get_all_attributes` ?

ima23: Thanks for checking & good point @ens-lgil! Currently this method reflects what REST endusers expect. I've created a jira note for revisiting the summary_as_hash method, as at the moment the attribute names in the db need to be revisited to make them as clear/clean/verbose as possible.

PR #198, opened 2018-07-11
Don't look up object id if already available

sarahhunt:


modules/Bio/EnsEMBL/Variation/PhenotypeFeature.pm
```diff
@@ -354,12 +354,10 @@ sub object {
...
-  
-  if (defined $object_id){
-    $self->{_object_id} = $object_id;
-  } elsif($self->object){
-    $self->{_object_id} =$self->object->stable_id()
-  }
+
+  $self->{_object_id} = $object_id if defined($object_id);
+
+  $self->{_object_id} ||= $self->object->stable_id();
```
ens-lgil: So it assumes that `$self->object` is always defined ?

ens-lgil:
Merged into `master` and `release/94`

PR #202, opened 2018-07-16
Feature/cv import

ima23:


scripts/import/import_clinvar_xml
```diff
@@ -56,18 +56,19 @@ use Data::Dumper;
 
 use Bio::EnsEMBL::Variation::Utils::Sequence qw( get_hgvs_alleles);
 use Bio::EnsEMBL::Variation::VariationFeature;
+use Bio::EnsEMBL::Slice;
 use Bio::DB::Fasta;
 use Bio::EnsEMBL::Registry;
 
-our $DEBUG = 0;
```
sarahhunt: One small thing - It's better to leave debug off by default

PR #204, opened 2018-07-23
Sift updates

at7:
Create a failure_reason table and add sift errors: Not enough sequences found by PSI-BLAST search, Less than the minimum number of sequences required.
Other problems that could be added: stack smashing detected (Could that be solved with more memory?), 100% identical with your protein query.
At the moment I'm not writing the table to file. This still needs to be done. Mouse runs very quickly and there is no benefit in discarding the md5s next release.

### DIFF
--- a/modules/Bio/EnsEMBL/Variation/Pipeline/ProteinFunction/RunSift.pm
+++ b/modules/Bio/EnsEMBL/Variation/Pipeline/ProteinFunction/RunSift.pm
@@ -138,7 +138,6 @@ sub run {
       }
       else {
         # If we just did not find enough hits, that's ok, just skip this
-        $alignment_ok = 0;
         my $error_file = $fasta_file . ".query.globalX.error";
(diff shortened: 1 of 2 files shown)


modules/Bio/EnsEMBL/Variation/Pipeline/ProteinFunction/RunSift.pm
```diff
@@ -138,20 +138,23 @@ sub run {
       }
       else {
         # If we just did not find enough hits, that's ok, just skip this
+        $alignment_ok = 0;
```
sarahhunt: Does this need setting? Doesn't everything in here either return or die?

at7: removed

at7:
I'll merge it to release/94 and master

PR #205, opened 2018-07-27
HGVS: don't report other inserted peptides after stop

sarahhunt:


ens-lgil:
Merged into `master` and `release/94`

PR #212, opened 2018-08-06
attrib and attrib_type updates

helensch:
Add new entries, remove old entry and tidy
- Replace empty descriptions with NULL
- Update to use consistent capitalisation


helensch:
The variation database creation still use attrib_entries.sql
https://www.ebi.ac.uk/seqdb/confluence/display/EV/Variation+database+creation and was used for this release

New databases and existing databases should have attribs updated using
 https://www.ebi.ac.uk/seqdb/confluence/display/EV/Attrib+tables+SOP

at7:
We should remove attrib_entries.sql from the variation repo and update the documentation to only use the production attrib tables. But I might oversee a good reason why we should keep the file.

helensch:
Removing the attrib_entries.sql would require access to the production attrib tables.
Suggest that attrib_entries.sql and generated from production attrib tables

at7:
Did you add the capitalisation updates to the admin site http://admin.ensembl.org/index.html too?

helensch:
Capitalisation was removed from attrib_entries.sql to make it consistent with what was in the production database. So no capitalisation updates were not done on the admin site

at7:
OK.

PR #221, opened 2018-09-04
Variation consequence - add rundates + tweak from memory

sarahhunt:


modules/Bio/EnsEMBL/Variation/Pipeline/FinishTranscriptEffect.pm
```diff
@@ -107,5 +109,20 @@ sub rejoin_table_files {
...
+
+  my $var_dba  = $self->get_species_adaptor('variation');
+
+  my $var_dbh = $var_dba->dbc->db_handle;
+
+  my $update_meta_sth = $var_dbh->prepare(qq[ insert ignore into meta
+                                              ( meta_key, meta_value) values (?,?)
+                                            ]);
+
+  $update_meta_sth->execute('TranscriptEffect_run_date', $self->run_date() );
```
ens-lgil: Is there a way to move this method "update_meta" in an other place in the API instead of rewriting it in several modules ?
This would be quite useful to use it in many places during production

PR #222, opened 2018-09-04
consider transcripts on reverse strand

at7:
Codons were not reverse complemented correctly for transcripts on the reverse strand. This has been fixed now and tested against variants overlapping transcripts on reverse and forward strand and compared to dbNSFP VEP tool results.

### DIFF
--- a/modules/Bio/EnsEMBL/Variation/Pipeline/ProteinFunction/RunDbNSFP.pm
+++ b/modules/Bio/EnsEMBL/Variation/Pipeline/ProteinFunction/RunDbNSFP.pm
@@ -92,7 +92,9 @@ sub run {
 
   my $translation = $self->get_translation($translation_stable_id);
   my $translation_seq = $translation->seq;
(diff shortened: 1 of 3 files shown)


modules/Bio/EnsEMBL/Variation/Pipeline/ProteinFunction/RunDbNSFP.pm
```diff
@@ -146,6 +148,7 @@ sub run {
         my $aaref = $data{'aaref'};
         my $mutated_triplet =  $new_triplets->{$triplet_seq}->{$nucleotide_position}->{$alt};
         my $mutated_aa = $codonTable->translate($mutated_triplet);
+        next if ($aaalt ne $mutated_aa);
```
ens-lgil: Confusing variable `$aaalt` (but this is an old variable) might miss to add `a`s. It would makes sense to name it `$alt_aa` like we have `$mutated_aa`.

ens-lgil:
Merged into `release/94` and `master`

PR #223, opened 2018-09-05
Fix for A//AA alleles

sarahhunt:


modules/Bio/EnsEMBL/Variation/DBSQL/VariationFeatureAdaptor.pm
```diff
@@ -1402,6 +1402,10 @@ sub _objs_from_sth {
               @clin_sig = split(/,/,$clin_sig );
             }
 
+            ## fix for e!94 allele strings with // or trailing /
+            $allele_string =~ s/\/\//\//;
+            $allele_string =~ s/\/$//;
```
ens-lgil: In this case ... maybe we could add the case where the '/' character is the first place ?
i.e. `$allele_string =~ s/^\///;`

sarahhunt: Thanks for the suggestion - I have checked and not found any of these.

PR #226, opened 2018-09-12
Ld 93 strand fix

at7:
We introduced looking up data from VCF files which don't have the strict_name_match flag set in the VCF config entry. This was introduced for farm species which have VCF files where the identifier is missing. I added the new code before the test if the VCF collection contained the required population which has been passed for the LD calculation. It then fails in the VCFCollection code which calls the seek method of the VCF parser when using a negative start value. I think Bio::DB::HTS should accept negative coordinates. But that would take longer to get updated because we are not in control of the repository. I also changed the order and check now first if the population is contained. Otherwise we don't need to worry about matching VCF variants to database variants.

### DIFF
--- a/modules/Bio/EnsEMBL/Variation/VCFCollection.pm
+++ b/modules/Bio/EnsEMBL/Variation/VCFCollection.pm
@@ -1487,6 +1487,7 @@ sub _seek {
   $self->_current($vcf);
   
   # now seek
+  $s = 0 if ($s < 0);
   $vcf->seek($c, $s, $e);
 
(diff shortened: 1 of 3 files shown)


modules/Bio/EnsEMBL/Variation/VCFCollection.pm
```diff
@@ -1487,6 +1487,7 @@ sub _seek {
   $self->_current($vcf);
   
   # now seek
+  $s = 0 if ($s < 0);
```
ens-lgil: Should it be `1` instead of `0` ?

at7:
Tests pass locally. Failure is because of an update on the DBI::mysql module. I will also create a PR for postreleasefix/94

PR #230, opened 2018-09-20
Store allele synonyms

helensch:


sql/table.sql
```diff
@@ -256,6 +256,29 @@ create table variation_synonym (
...
+*/
+
+CREATE TABLE allele_synonym (
+  allele_synonym_id int(10) unsigned NOT NULL AUTO_INCREMENT,
+  variation_id      int(10) unsigned NOT NULL,
+  hgvs              varchar(600) NOT NULL,
+  name              varchar(255) NOT NULL,
+  PRIMARY KEY (allele_synonym_id),
+  UNIQUE KEY variation_name_idx (variation_id, name),
+  KEY hgvs_idx (hgvs)
```
sarahhunt: We also need an index on the name to enable look up by CA id.

sql/table.sql
```diff
@@ -256,6 +256,29 @@ create table variation_synonym (
...
+@column hgvs                  HGVS of variant e.g. NC_000001.11:g.230710048A>G
+@column name                  Name of the allele synonym e.g. CA127784
+
+@see variation
+*/
+
+CREATE TABLE allele_synonym (
+  allele_synonym_id int(10) unsigned NOT NULL AUTO_INCREMENT,
+  variation_id      int(10) unsigned NOT NULL,
+  hgvs              varchar(600) NOT NULL,
```
sarahhunt: We refer to this value as hgvs_genomic in the transcript_variation table; it may be useful to be consistent.

sarahhunt:
It is customary to use patch 'a' for the schema version change. I don't think anything is dependent on this.

sql/table.sql
helensch: Is the only genomic hgvs going to be used?

sql/table.sql
helensch: I'll add  that.  Should I remove the hgvs index?

helensch:
Do you want me to rename this to patch b?
I'll add the patch to the table.sql

sql/table.sql
sarahhunt: I can't currently think of a reason to use any other.

sql/table.sql
sarahhunt: Do you need it for populating the variant summary info on the website?

sarahhunt:
It may be clearer to call this patch b.

sql/table.sql
helensch: I'll change the name. The ClinGen Allele Registry supports search using HGVS expressions for a transcripts, genomic and protein sequences .

helensch:
Replaced by #231 that includes schema patch and addition of allele_synonym table.

PR #236, opened 2018-10-04
use gene symbol if present, otherwise use gene stable id

ima23:


ens-lgil:

PR #237, opened 2018-10-12
fixes alias in delete sql command

manuelcarbajo:
Fixes the truncate sql command in the analysis InitJobs. This command is only used when SIFT is run on full (instead of adding extra SIFTs) which is not done so often. This probably explains why this has gone unnoticed for a while. I have tested the fix and the pipeline passes beyond the analysis.


ens-lgil:
Hi Manuel,

Please, could you resubmit your PR to the branch `postreleasefix/95` ?

Thanks,
Laurent

manuelcarbajo:
Moved pull request to postreleasefix/95
https://github.com/Ensembl/ensembl-variation/pull/238

PR #238, opened 2018-10-12
Postreleasefix/95

manuelcarbajo:
Fixes the truncate sql command in the analysis InitJobs. This command is only used when SIFT is run on full (instead of adding extra SIFTs) which is not done so often. This probably explains why this has gone unnoticed for a while. I have tested the fix and the pipeline passes beyond the analysis.


ens-lgil:
Sorry Manuel, but this PR is a bit messy.
Please, could you resent your PR from a cleaned `postreleasefix/95` branch ?

Thanks,
Laurent

PR #239, opened 2018-10-12
Fixes the truncate sql command in fetch_input()

manuelcarbajo:
Fixes the truncate sql command in the analysis InitJobs. This command is only used when SIFT is run on full (instead of adding extra SIFTs) which is not done so often.  I have tested the fix and the pipeline passes beyond the analysis.


ens-lgil:
@manuelcarbajo,

The pull request has been done from the main repo and not your fork.
We tend to avoid doing that within the Variation team and in your case you were able to do so because you have admin rights on the whole Project.
I am going to close this PR and I will create a new one with your suggested changes.
Thanks,
Laurent

PR #240, opened 2018-10-15
Fixes the truncate sql command in fetch_input() - from Manuel Carbajo

ens-lgil:


ens-lgil:
Merged into `release/95`and `master`

PR #241, opened 2018-10-18
protein function  parameter used in InitJobs L240

ima23:


helensch:
As the code in InitJobs.pm that uses the param('use_compara') is checking sift_protein_db_version
https://github.com/Ensembl/ensembl-variation/blob/postreleasefix/95/modules/Bio/EnsEMBL/Variation/Pipeline/ProteinFunction/InitJobs.pm#L240
Suggest changing the param in InitJobs.pm to param('sift_use_compara') and set this parameter in
ProteinFunction_conf.pm for the init_jobs analysis. This would allow  allow polyphen_use_compara to be used for polyphen specific settings in InitJobs.pm if needed.

ima23:
Thanks  for the added link to L240 @helensch. 

If in the future InitJobs.pm will make use of the parameter `pph_use_compara`, then yes, it would make sense to change the `InitJobs.pm` currently expected parameter 'use_compara' to the more specific version `sift_use_compara`. This would need 2 changes, one in the `ProteinFunction_conf.pm` and `InitJobs.pm`.

The current correction is the minimal change, 1 change to `ProteinFunction_conf.pm` to conform to what `InitJobs.pm` expects. Unless there are concrete plans for more work on `InitJobs.pm` that would benefit from more changes now, I am leaning in favour for the minimal code change.

helensch:
Current InitJobs.pm makes a distinction between SIFT and PolyPhen parameters that is why I suggested the distinction. If you want to make mininal code changes thats fine.

ens-lgil:

PR #242, opened 2018-10-18
small log message fixes for import script

ima23:


scripts/import/import_clinvar_xml
```diff
@@ -350,7 +351,7 @@ sub get_feature{
    }
 
    ## if multiple rsIDs for the same ReferenceClinVarAssertion print warning
-   warn "multiple rsIDs for same RCV: $accession, rsIDs: ".join(",",@{$feature{measureXrefs}{$measure->{ID}}{dbSNP}{rs}} ) if (defined $feature{measureXrefs}{$measure->{ID}}{dbSNP} && scalar @{$feature{measureXrefs}{$measure->{ID}}{dbSNP}{rs}} >1 );
+   warn "multiple rsIDs for same RCV: $accession, rsIDs: ".join(",",@{$feature{measureXrefs}{$measure->{ID}}{dbSNP}{rs}} ) if (defined $feature{measureXrefs}{$measure->{ID}}{dbSNP} && scalar @{$feature{measureXrefs}{$measure->{ID}}{dbSNP}{rs}} >1 )."\n";
```
sarahhunt: The '\n' looks as if its in the wrong place - has this been tested with some warning-triggering data?

ima23: Thanks for spotting that. This warning was fixed after the import by looking at the logs that triggered it.

PR #243, opened 2018-10-24
Standardise table.sql file

james-monkeyshines:
## Description and Use case
The new datacheck for validating a db schema uses the table.sql file.
This is mostly consistent already, both within the file and between instances of the files in different repositories. But there are a few simple changes to terminology and formatting that would further standardise the files. These sorts of things can be handled in the datacheck code, but having fewer, and less complicated, exceptions makes the code more robust and easier to understand.

## Benefits
Datacheck is simpler. Standardisation and consistency are generally good things to aim for.

## Possible Drawbacks
None I can think of.

## Testing
Unit tests not relevant.
I created databases before and after the changes, and the results were identical.

### DIFF
--- a/sql/table.sql
+++ b/sql/table.sql
@@ -76,7 +76,7 @@ CREATE TABLE variation (
   display INT(1) DEFAULT 1,
 
 	PRIMARY KEY( variation_id ),
-	UNIQUE ( name ),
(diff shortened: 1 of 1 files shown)


james-monkeyshines:
Thanks. Yes, this is for e!96.

PR #244, opened 2018-10-25
Regulation updates 95

at7:
Improve regulation pipeline by using split slice and calculate overlap between variant and regulatory or motif feature from split slice. Add option for using experimentally validated motif feature only (only works for human).


at7:

PR #245, opened 2018-10-25
Remapping updates 95

at7:
Move DBC getter methods to base module. Hopefully, this will avoid confusion between connections to old and new databases.


ens-lgil:

PR #247, opened 2018-10-25
remapping code for VCF files

at7:
Pipeline for remapping VCF files against a remapped variation database. The pipeline was tested for cow and horse.


at7:

PR #248, opened 2018-10-26
Fix attrib count and open file check

helensch:
Reports for all attrib_id. In future need to limit this to co-located allele.
Reports if file open fails.


ens-lgil:

PR #251, opened 2018-10-31
update to remapped files for cow and horse

at7:


ens-lgil:

PR #254, opened 2018-11-09
Genome wide scores prf95

at7:
Suggestions for adding CADD and GERP scores from file. At the moment scores are only returned by variation feature.


.travis.yml
```diff
@@ -34,6 +34,9 @@ before_install:
 - cd C_code/
 - make
 - cd ../
+- wget https://github.com/ucscGenomeBrowser/kent/archive/v335_base.tar.gz
```
ens-lgil: Wouldn't be better to create a `get_dependencies.sh` file like we have for VEP instead ?
https://github.com/Ensembl/ensembl-vep/blob/release/95/travisci/get_dependencies.sh
and
https://github.com/Ensembl/ensembl-vep/blob/release/95/.travis.yml#L43

PR #255, opened 2018-11-09
Feature/spdi - new method 'fetch_by_spdi_notation'

dglemos:
Method that parses a SPDI notation: 
S:P:D:I (SequenceId:Position:DeletedSequence:InsertedSequence) 

This method supports substitutions, deletions, insertions and indels.


modules/Bio/EnsEMBL/Variation/DBSQL/VariationFeatureAdaptor.pm
```diff
@@ -2348,4 +2348,192 @@ sub fetch_all_by_location_identifier {
...
+  ########################### Check & split input ########################### 
+  my ($sequence_id, $position, $deleted_seq, $inserted_seq) = split /:/, $spdi; 
+
+  # strip version number from reference 
+  if($sequence_id =~ m/\./i){
+    $sequence_id =~ s/\.\d+//g;
+  } 
+
+  # count number of fields  
+  my $count_separator = () = $spdi =~ m/\:/g;
```
ens-lgil: Maybe you could use the`split` in line 2377 in a way to avoid recounting the fields

modules/Bio/EnsEMBL/Variation/DBSQL/VariationFeatureAdaptor.pm
```diff
@@ -2348,4 +2348,192 @@ sub fetch_all_by_location_identifier {
...
+    else{
+      $ref_allele = $refseq_allele;    
+    }
+    $alt_allele = uc $inserted_seq; 
+    throw ("Reference allele given by SPDI notation $spdi ($ref_allele) matches alt allele given by SPDI notation $spdi ($alt_allele)")
+      unless ($ref_allele ne $alt_allele);  
+  }  
+
+  # Variation is a deletion
+  elsif($deleted_seq =~ m/^([A-Z]+|\d+)$/i && $inserted_seq eq ''){
```
ens-lgil: You could replace the regex `m/^([A-Z]+|\d+)$/i` by simply `m/^\w+$/`

modules/Bio/EnsEMBL/Variation/DBSQL/VariationFeatureAdaptor.pm
```diff
@@ -2348,4 +2348,192 @@ sub fetch_all_by_location_identifier {
...
+  my ($start, $end, $strand, $ref_allele, $alt_allele); 
+
+  $strand = 1; ## strand should be genome strand for SPDI genomic notation
+
+  # Get a slice adaptor to enable check of supplied reference allele
+  my $slice_adaptor = $user_slice_adaptor || $self->db()->dnadb()->get_SliceAdaptor(); 
+  my $slice = $slice_adaptor->fetch_by_region('chromosome', $sequence_id ) || $slice_adaptor->fetch_by_region(undef, $sequence_id);  
+
+  # Variation is a substitution  
+  if($deleted_seq =~ m/^([A-Z]+|\d+)$/i && $inserted_seq =~ m/^([A-Z]+)$/i && length($deleted_seq) == length($inserted_seq) && length($deleted_seq) == 1){
```
ens-lgil: You could replace the regex m/^([A-Z]+|\d+)$/i by simply m/^\w+$/

modules/Bio/EnsEMBL/Variation/DBSQL/VariationFeatureAdaptor.pm
```diff
@@ -2348,4 +2348,192 @@ sub fetch_all_by_location_identifier {
...
+  my $slice = $slice_adaptor->fetch_by_region('chromosome', $sequence_id ) || $slice_adaptor->fetch_by_region(undef, $sequence_id);  
+
+  # Variation is a substitution  
+  if($deleted_seq =~ m/^([A-Z]+|\d+)$/i && $inserted_seq =~ m/^([A-Z]+)$/i && length($deleted_seq) == length($inserted_seq) && length($deleted_seq) == 1){ 
+    $start = $position + 1; 
+    $end = $start;  
+
+    my $refseq_allele = get_reference_allele($slice_adaptor, $sequence_id, $start, $end); 
+
+    if($deleted_seq =~ m/^([A-Z]+)$/i){
```
ens-lgil: Just a bit concerned by all the repeated regexs to check `$deleted_seq` and `$insertion_seq`.
For maintenance, it might be easier to store this match information in variables playing a role as flags.

modules/Bio/EnsEMBL/Variation/DBSQL/VariationFeatureAdaptor.pm
```diff
@@ -2348,4 +2348,192 @@ sub fetch_all_by_location_identifier {
...
+
+  #######################  extract genomic coordinates and reference seq allele  ####################### 
+
+  my ($start, $end, $strand, $ref_allele, $alt_allele); 
+
+  $strand = 1; ## strand should be genome strand for SPDI genomic notation
+
+  # Get a slice adaptor to enable check of supplied reference allele
+  my $slice_adaptor = $user_slice_adaptor || $self->db()->dnadb()->get_SliceAdaptor(); 
+  my $slice = $slice_adaptor->fetch_by_region('chromosome', $sequence_id ) || $slice_adaptor->fetch_by_region(undef, $sequence_id);
```
ens-lgil: More general question: is spdi it only meant for notations on chromosome ?

sarahhunt: No - we are starting with the simplest use case but will need to support Ensembl and RefSeq transcripts eventually.

modules/Bio/EnsEMBL/Variation/DBSQL/VariationFeatureAdaptor.pm
```diff
@@ -2348,4 +2348,196 @@ sub fetch_all_by_location_identifier {
...
```
sarahhunt: It would be better to check a slice is returned and throw if not. Otherwise, a error is returned:

Can't call method "seq" on an undefined value at ensembl-variation/modules/Bio/EnsEMBL/Variation/DBSQL/VariationFeatureAdaptor.pm line 2497

modules/Bio/EnsEMBL/Variation/DBSQL/VariationFeatureAdaptor.pm
sarahhunt: We try report input errors clearly to avoid anyone using the API's having to dig into our code to work out what is going on.

If I enter 'NC_000022.11:25459491:G:C:C' I get a message : 

MSG: Could not parse the SPDI notation NC_000022.11:25459491:G:C:C

which is helpful. (It might be even more helpful to day something like 'too many elements present') If I enter 'NC_000077.14:97795919:T:C' I get the message:

MSG: Bio::EnsEMBL::Slice arg expected

which isn't so clear. People may get in touch thinking they have found a bug, when there is just a typo in the sequence name.

PR #263, opened 2018-11-15
Feature/spdi genomic

dglemos:
This method returns a SPDI notation for a VariationFeature. 
In a first stage, it doesn't take any inputs. It uses the slice it is placed on as the reference.

### DIFF
--- a/modules/Bio/EnsEMBL/Variation/VariationFeature.pm
+++ b/modules/Bio/EnsEMBL/Variation/VariationFeature.pm
@@ -1968,7 +1968,85 @@ sub hgvs_genomic {
 
 }
 
+=head2  spdi_genomic
 
+  Description : Returns a reference to a hash with the allele as key and a string with the genomic SPDI notation of this VariationFeature as value. By default uses the
+                slice it is placed on as reference but a different reference feature can be supplied.
+  Returntype  : Hash reference
+  Exceptions  : Throws exception if VariationFeature can not be described relative to the feature_Slice of the supplied reference feature
+  Caller      : general
+  Status      : Experimental
+
+=cut
+
+sub spdi_genomic{ 
+
+  my $self = shift; 
+  
+  my $ref_feature = $self;   
+  my %spdi; 
+
+  my $ref_slice = $self->slice; 
+
+  my $tr_vf = $self; 
+  my ($vf_start, $vf_end, $ref_length) = ($tr_vf->start, $tr_vf->end, ($ref_feature->end - $ref_feature->start) + 1);
+
+  return {} if ($vf_start < 1 || $vf_end < 1);  
+
+  my $vf_strand = $self->strand(); 
+  
+  # set up sequence reference 
+  my $syn = $ref_slice->get_all_synonyms('RefSeq_genomic'); 
+  my $reference_name = (defined $syn->[0] ? $syn->[0]->name() : $ref_feature->seq_region_name()); 
+
+  my @all_alleles = split(/\//,$tr_vf->allele_string());
+  my $ref_allele = shift @all_alleles; 
+  my $spdi_ref_allele; 
+  my $spdi_alt_allele; 
+  
+  # Create a spdi notation for each allele 
+  foreach my $alt_allele (@all_alleles){ 
+    
+    # Expand tandems before check for non nucleotide character 
+    expand(\$alt_allele); 
+
+    # Skip if the allele contains weird characters 
+    next if $alt_allele =~ m/[^ACGT\-]/ig; 
+
+    my $chr_start = $tr_vf->start;
+    my $chr_end   = $tr_vf->end; 
+
+    # If the variant is a substitution, deletion or indel 
+    my $spdi_position = $chr_start - 1;  
+
+    if($chr_start == $chr_end){ 
+      $spdi_ref_allele = $ref_allele; 
+      
+      # Variation is a deletion  
+      if($alt_allele eq '-'){ $spdi_alt_allele = ""; }  
+
+      # Variation is a substitution 
+      else{                   $spdi_alt_allele = $alt_allele; }
+
+    } 
+    # Variation is an insertion 
+    elsif($ref_allele eq '-'){ 
+      $spdi_position = $chr_end; 
+      $spdi_ref_allele = "";  
+      $spdi_alt_allele = $alt_allele; 
+    }
+    # Variation is an indel  
+    else{
+      $spdi_ref_allele = $ref_allele;   
+      $spdi_alt_allele = $alt_allele; 
+    }
+ 
+    my $spdi_notation = $reference_name . ":" . $spdi_position . ":" . $spdi_ref_allele . ":" . $spdi_alt_allele; 
+    $spdi{$alt_allele} = $spdi_notation; 
+  }
+
+  return \%spdi; 
+} 
 
 sub length {
   my $self = shift;
--- a/modules/Bio/EnsEMBL/Variation/VariationFeature.pm
+++ b/modules/Bio/EnsEMBL/Variation/VariationFeature.pm
@@ -1971,7 +1971,7 @@ sub hgvs_genomic {
 =head2  spdi_genomic
 
   Description : Returns a reference to a hash with the allele as key and a string with the genomic SPDI notation of this VariationFeature as value. By default uses the
-                slice it is placed on as reference but a different reference feature can be supplied.
+                slice it is placed on as reference. 
   Returntype  : Hash reference
   Exceptions  : Throws exception if VariationFeature can not be described relative to the feature_Slice of the supplied reference feature
   Caller      : general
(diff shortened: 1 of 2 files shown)


modules/Bio/EnsEMBL/Variation/VariationFeature.pm
```diff
@@ -1968,7 +1967,84 @@ sub hgvs_genomic {
...
+  my $ref_feature = $self;   
+  my %spdi; 
+
+  my $ref_slice = $self->slice; 
+ 
+  my ($vf_start, $vf_end, $ref_length) = ($ref_feature->start, $ref_feature->end, ($ref_feature->end - $ref_feature->start) + 1);
+
+  return {} if ($vf_start < 1 || $vf_end < 1);  
+
+  my $vf_strand = $self->strand();
```
helensch: Is $vf_strand used?

modules/Bio/EnsEMBL/Variation/VariationFeature.pm
```diff
@@ -1968,7 +1967,84 @@ sub hgvs_genomic {
...
+    my $spdi_position = $chr_start - 1;  
+
+    if($chr_start == $chr_end){ 
+      $spdi_ref_allele = $ref_allele; 
+      
+      # Variation is a deletion  
+      if($alt_allele eq '-'){ $spdi_alt_allele = ""; }  
+
+      # Variation is a substitution 
+      else{                   $spdi_alt_allele = $alt_allele; }
```
helensch: Can indentation be fixed?

modules/Bio/EnsEMBL/Variation/VariationFeature.pm
```diff
@@ -1968,7 +1967,84 @@ sub hgvs_genomic {
...
+  Caller      : general
+  Status      : Experimental
+
+=cut
+
+sub spdi_genomic{ 
+
+  my $self = shift; 
+  
+  my $ref_feature = $self;
```
sarahhunt: It's not necessary to rename self here.

modules/Bio/EnsEMBL/Variation/VariationFeature.pm
```diff
@@ -1968,7 +1967,84 @@ sub hgvs_genomic {
...
+
+    } 
+    # Variation is an insertion 
+    elsif($ref_allele eq '-'){ 
+      $spdi_position = $chr_end; 
+      $spdi_ref_allele = "";  
+      $spdi_alt_allele = $alt_allele; 
+    }
+    # Variation is an indel  
+    else{
```
sarahhunt: This is not capturing longer deletions correctly;  rs371609562    (TT/- )   returns NC_000011.9:61395:TT:-  

The check for ($alt_allele eq '-') should be outside the if($chr_start == $chr_end)

modules/Bio/EnsEMBL/Variation/VariationFeature.pm
```diff
@@ -1968,7 +1967,84 @@ sub hgvs_genomic {
...
+  my $syn = $ref_slice->get_all_synonyms('RefSeq_genomic'); 
+  my $reference_name = (defined $syn->[0] ? $syn->[0]->name() : $ref_feature->seq_region_name()); 
+
+  my @all_alleles = split(/\//,$ref_feature->allele_string());
+  my $ref_allele = shift @all_alleles; 
+  my $spdi_ref_allele; 
+  my $spdi_alt_allele; 
+  
+  # Create a spdi notation for each allele 
+  foreach my $alt_allele (@all_alleles){
```
sarahhunt: I know it wasn't in the original spec, but it would be useful to optionally return the reference spdi. We could have a boolean (include_reference_allele?) optionally passed in to turn this on and it could default to 0 ( alt alleles only) without.

modules/Bio/EnsEMBL/Variation/VariationFeature.pm
```diff
@@ -1968,7 +1967,84 @@ sub hgvs_genomic {
...
+=cut
+
+sub spdi_genomic{ 
+
+  my $self = shift; 
+  
+  my $ref_feature = $self;   
+  my %spdi; 
+
+  my $ref_slice = $self->slice;
```
sarahhunt: As VariationFeatures can be created without slices, it would be better to check we have something here and throw if not.

modules/Bio/EnsEMBL/Variation/VariationFeature.pm
```diff
@@ -1968,7 +1967,117 @@ sub hgvs_genomic {
 
 }
 
+=head2  spdi_genomic
+
+Arg [1]     : int (Optional)
+              It also returns a SPDI notation for the reference allele. By default value is '0'.
+              '1' -> returns a SPDI for the reference allele;
+              '0' -> doesn't return a SPDI for the reference allele;
+Example     : my $vf = $variation_feature_adaptor->fetch_by_dbID(rs145160881);
```
helensch: parameter to fetch_by_dbID should be a variation_feature_id or a  "location_identifier"
Suggest say "For example" instead of will print

modules/Bio/EnsEMBL/Variation/VariationFeature.pm
```diff
@@ -1968,7 +1967,117 @@ sub hgvs_genomic {
...
+
+sub spdi_genomic{
+
+  my $ref_feature = shift;
+  my $include_ref_allele = shift;
+
+  my %spdi;
+  if(!$include_ref_allele){ $include_ref_allele = 0; }
+  # If input is provided, it must have value '1' (include reference allele) or '0' (not include reference allele)
+  throw("Include reference allele must be a numeric value '1' or '0'.") unless ($include_ref_allele =~ m/^(1|0)/i);
```
helensch: Pattern match allows any value starting 0 or 1. Case-insensitve not needed. Suggest =~ /^[01]$/ or use numerical comparison

modules/Bio/EnsEMBL/Variation/VariationFeature.pm
```diff
@@ -1968,7 +1967,118 @@ sub hgvs_genomic {
...
+      $spdi_ref_allele = $ref_size unless ( $ref_size <= 20 );
+      $spdi_alt_allele = "";
+    }
+    # Variation is an indel
+    else{
+      $spdi_ref_allele = $ref_allele;
+      $spdi_alt_allele = $alt_allele;
+    }
+
+    reverse_comp(\$spdi_ref_allele) if( $flip_allele == 1 && $spdi_ref_allele && $spdi_ref_allele =~ m/^[A-Z]+/i);
```
helensch: Not sure why pattern match A-Z?

dglemos: Sometimes the ref_allele can be a number, for example a deletion of 21 bp (NC_000002.12:47404148:10:) in this case the reverse would be 12. So, I put the pattern to make sure it would only reverse sequences with letters.

helensch: It might be useful to add a comment about this, as earlier only pattern match  $alt_allele =~ m/[^ACGT\-]/ig;
What is returned if need to reverse on longer deletions?
Should the reverse_comp of ref and alt be grouped in an if (flip_allele == 1)

modules/t/hgvs_parser.t
```diff
@@ -698,6 +698,15 @@ sub test_output{
   ok(  $hgvs_genomic->{$allele} eq $output->[0], "$input genomic level $output->[0], $output->[5]" );
   if($DEBUG==1){print "TEMP: $input => gen; expected $output->[0]\t returns: $hgvs_genomic->{$allele} for allele:$allele\n";} 
 
+  # hgvs genomic with slice and without a slice
```
helensch: This test gets done several times. It should be run outside this sub and only once.

modules/t/test-genome-DBs/homo_sapiens/variation/variation_feature.txt
```diff
@@ -1295,3 +1295,6 @@
 111657317	27508	45420082	45420082	1	112643726	G/T	rs7569588	1	\N	1	non_coding_transcript_variant,intron_variant,upstream_gene_variant	2,4,5,6,7,9,10,11,12,13,14,17,18,19,20,21,22,42,43,44,45,47,48,49,50,51,52,53,54	2	0	T	0.481629	2412	1	\N	367,368,369,370	1
 47738282	100965601	6721	6721	1	48717434	T/C	rs199476127	1	\N	1	intergenic_variant		0	0	\N	\N	\N	1	\N	\N	1
 997738282	100965601	6721	6721	1	48717434	T/C	rs199476127	1	\N	1			0	0	\N	\N	\N	1	\N	\N	1
+997738292	27526	48332373	48332376	-1	9971382	AAAT/-	rs1480172069	1	\N	1	3_prime_UTR_variant		12	0	\N	\N	\N	1	\N	\N	1
```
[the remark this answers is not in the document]
dglemos: yes, they are already there.

PR #264, opened 2018-11-20
annotate remaining features on split_slice

at7:


ens-lgil:
Merged into `release/95` and `master`

PR #269, opened 2018-12-12
Revamp the script generating the Population web documentation to use …

ens-lgil:
…less hard coded data and more information from the vcf_config file and the Variation databases

### DIFF
--- a/scripts/misc/generate_population_table.pl
+++ b/scripts/misc/generate_population_table.pl
@@ -69,98 +69,36 @@ my $json_string = <IN>;
 # parse JSON into hashref $config
 my $vcf_config = JSON->new->decode($json_string) or throw("ERROR: Failed to parse config file $vcf_config_file");
 
-my %pops_list;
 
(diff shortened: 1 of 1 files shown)


scripts/misc/generate_population_table.pl
```diff
@@ -410,55 +369,197 @@ sub get_size {
...
+# Mostly based on the vcf_config file and Variation databases
 sub get_project_populations {
-  my $projects = shift;
-  
+
   foreach my $project (@{$vcf_config->{'collections'}}) {
     my $project_id = $project->{'id'};
-    my $project_prefix = ($project->{'population_prefix'}) ? $project->{'population_prefix'} : '';
-    if ($projects->{$project_id}) {
+    next if ($project_id =~ /GRCh37/i);
```
at7: Why are you not using the assembly field for filtering by assembly?

ens-lgil: Good point!

PR #274, opened 2018-12-18
Fix schema version 95

helensch:


at7:
Irina's patch didn't update the variation meta table: https://github.com/Ensembl/ensembl-variation/pull/233 But core and funcgen have been updated. The allele_synonym patch went in though. Something to look out for when we patch test dbs for 96.

at7:
Merged to release/95

PR #277, opened 2018-12-18
Task/motif name update 96

at7:
This looks bigger than it is. It really is just renaming the motif_name column to binding_matrix_stable_id in the motif_feature_variation table. https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-1444

### DIFF
--- a/modules/Bio/EnsEMBL/Variation/MotifFeatureVariation.pm
+++ b/modules/Bio/EnsEMBL/Variation/MotifFeatureVariation.pm
@@ -121,6 +121,29 @@ sub motif_name {
   return $self->{motif_name};
 }
 
+=head2 binding_matrix_stable_id
+
+  Arg [1]    : string $stable_id (optional)
+               The new value to set the binding_matrix_stable_id attribute to
+  Example    : $binding_matrix_stable_id = $motif_feature_variation->binding_matrix_stable_id()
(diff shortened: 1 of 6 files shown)


sql/table.sql
```diff
@@ -1536,7 +1536,7 @@ CREATE TABLE IF NOT EXISTS motif_feature_variation (
                                           'TFBS_amplification',
                                           'TFBS_translocation'
                                         ),
-    motif_name                          VARCHAR(60) DEFAULT NULL,
+    binding_matrix                      VARCHAR(60) DEFAULT NULL,
```
ens-lgil: Should it rather be `binding_matrix_stable_id` ?

modules/Bio/EnsEMBL/Variation/MotifFeatureVariation.pm
```diff
@@ -121,6 +121,29 @@ sub motif_name {
   return $self->{motif_name};
 }
 
+=head2 binding_matrix_stable_id
```
ens-lgil: Should we deprecate the method `motif_name` then ?

at7: No. It is still used.

ens-lgil: Oups! Sorry I didn't read the content of the method! My bad!

PR #279, opened 2018-12-18
add release/95 funcgen patches

at7:
These are missing funcgen 95 patches. I most likely forgot to add them after funcgen's late branching. They need to go into master before we patch for 96 tomorrow.


ens-lgil:
Ok, let's wait for the tests to finish before merging it

at7:
Merged to master and release/95

PR #283, opened 2019-01-03
Fix small issues in fetch_by_spdi_notation and added new SPDI format 

dglemos:


helensch:
Merged into master  and release/96

PR #285, opened 2019-01-10
Fix HTML file name change in variant web doc

ens-lgil:


ima23:
@aparton was this merged to master and release branch?

PR #290, opened 2019-01-17
add dump files for phenotypes plugin script

ima23:
Script hast the option to dump phenotypes for all variation dbs for a given release from a database host, or dump only specific species.

`perl dump_phenotypes.pl -output_dir $outDir -v 95 -hname $hostName -u $user`

or

`perl dump_phenotypes.pl -output_dir $outDir -v 95 -hname $hostName -u $user -species sus_scrofa,gallus_gallus`


ima23:
@ens-lgil: good suggestion! I'll move it to `export`. 
There are a few other `dump` scripts in import folder, I'll leave them there, but can move them while I'm there if you think it might be good.

PR #292, opened 2019-01-21
Fix strange characters for Orphanet and GWAS imports

ens-lgil:


ima23:
Merged to `release/96` and `master`

PR #293, opened 2019-01-22
add refseq option parameter to variant simulator script

ima23:


tools/variant_simulator/simulate_variation
```diff
@@ -77,9 +78,13 @@ print "getting species ($species)";
 print ", chrom ($chromOnly)" if $chromOnly;
 print ", gene ($gene)", if $gene;
 print ", out: $outFile\n";
+my $dbtype = "core";
+$dbtype = "otherfeatures" if $refSeq_genes;
```
ens-lgil: Could we use a more condensed syntax for it ?
``` $dbtype = $refSeq_genes ?  "otherfeatures" : "core";```

File: tools/variant_simulator/simulate_variation [hunk 2]
```diff
@@ -77,9 +78,13 @@ print "getting species ($species)";
...
+my $logic_name;
+$logic_name = "refseq_import" if $refSeq_genes;
```
ens-lgil: Same as above:
```$logic_name = $refSeq_genes ? "refseq_import" : undef;```

File: tools/variant_simulator/simulate_variation [hunk 3]
```diff
@@ -49,6 +49,7 @@ GetOptions( "species=s" => \$species,
     "codingOnly" => \$codingOnly,
     "intronsOnly" => \$intronOnly,
     "assembly=s" => \$assembly,
+    "refseq"     => \$refSeq_genes,
```
ens-lgil: For consistency, could we use the same case for `refseq` and `$refSeq_genes` ?
Usually `refseq` is written in Ensembl in lower case.

ima23: Good point! I've updated this in the next commit.

tools/variant_simulator/simulate_variation
ima23: Thanks for the suggestion! I've updated this in the next commit.

tools/variant_simulator/simulate_variation
ima23: Now updated.

PR #294, opened 2019-01-22
A couple of import script updates

ens-lgil:


scripts/import/import_all_sv_data.pl
```diff
@@ -1351,12 +1353,41 @@ sub parse_line {
...
 
+  $info->{chr} = ($seq_regions{$info->{chr}}) ? $seq_regions{$info->{chr}} : get_seq_region($info->{chr});
+
   $info = parse_9th_col($info,\@last_col);
 
   return $info;
 }
 
 
+sub get_seq_region {
```
at7: Is there no API support for retrieving the seq_region_name synonyms? What happens if no synonym was found?

ens-lgil: The script doesn't use much the API (mainly SQL queries).
I need to set something if no seq_region are found.
This edit was done because of the non usual seq regions for Chicken

PR #297, opened 2019-01-25
Update script import variant synonyms 

dglemos:


scripts/import/import_variant_synonyms
```diff
@@ -202,6 +266,66 @@ sub get_source{
...
+
+  my $dest_db = shift;
+  my $prod_dbc = shift; 
+
+  # my %source_names; # hash with source id and corresponding source name 
+  my %source_list; 
+
+  my $source_adaptor = $dest_db->get_SourceAdaptor('human', 'variation'); 
+
+  my $source_output = query_synonyms_data($prod_dbc, '*', 'source');
```
sarahhunt: Could you specify the columns here to ensure they are returned in the expected order? We have had problems when patches changed the default order returned in the past.

scripts/import/import_variant_synonyms
```diff
@@ -45,50 +45,83 @@ use Bio::EnsEMBL::Variation::Source;
...
+  my $prod_dbc = Bio::EnsEMBL::DBSQL::DBConnection->new
+    ( '-host'    => $host,
+      '-port'    => $port, 
+      '-user'    => $user,
+      '-pass'    => $pass,
+      '-dbname'  => $db_name, 
+      '-driver'  =>  'mysql' 
+    ); 
+
+  my $synonyms_output = query_synonyms_data($prod_dbc, '*', 'human_synonym');
```
sarahhunt: Could you specify the columns here to ensure they are returned in the expected order?

PR #300, opened 2019-01-30
Added correct annotation of inframe_insertion variants

aparton:
Roundabout fix for https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-1140

inframe_insertion and inframe_deletion already exist - uses inframe insertions properly


sarahhunt:
This would classify rs138498319 HGVSp=ENSP00000402674.2:p.Ser462delinsLysLeuArg as an inframe_insertion. It could be argued that at genomic level it is an insertion, but at protein level it is more complex, so the classification is misleading.

dbSNP uses inframe_indel  here:
https://www.ncbi.nlm.nih.gov/snp/rs138498319

which is less open to misinterpretation.

PR #301, opened 2019-01-31
updated BAM for MANE 0.5 release

sarahhunt:


ens-lgil:
Is this for e97 ?

sarahhunt:
Yes - the matching GFF should be going into the otherfeatures database for e!96, so we will run SIFT and PolyPhen2 for e!97.

PR #305, opened 2019-02-13
Update the BioPerl directory

ens-lgil:


at7:
So what happened here? Has bioperl removed a release branch?

ens-lgil:
This is related to this PR in ensembl-production: https://github.com/Ensembl/ensembl-production/pull/214
Basically the BioPerl path we put in PERL5LIB was wrong (outdated) so Bio-DB-HTS didn't find the BioPerl installation , therefore it tried to download the BioPerl version `1.7.2`, which is quite different from the `1.6.924` and it caused an issue with the rest of the installation

at7:
That makes sense now. Thank you!

ens-lgil:
Wrong commit

PR #306, opened 2019-02-13
Fix path to BioPerl in the travis PERL5LIB

ens-lgil:
Cleaned version of the PR #305


ens-lgil:
The file acces right to travisci/harness.sh was changed and it wasn't expected

PR #307, opened 2019-02-14
replace empty string by NULL

at7:
This came up in our health checks. Empty strings are now replaced by NULL.


ens-lgil:

PR #308, opened 2019-02-14
Fix path to BioPerl in the travis PERL5LIB

ens-lgil:
Same as #306 but without the read/write access change


at7:
merged to master and release/96

PR #310, opened 2019-02-14
Yearly copyright update - 2019

ens-lgil:
Only merge it with the branch `release/96` to avoid overwritting some commits only pushed on  the `master` branch.


ima23:
merged with release/96

PR #312, opened 2019-02-15
Add support for UniProt accessions in HGVSp input

ens-lgil:


sarahhunt:
This looks good, thanks!  One small thing - in VEP, it is triggering a message saying:

"WARNING: Possible invalid use of gene name 'P01019' as HGVS reference; most likely transcript will be selected"

changing this to 'gene or protein identifier' would be clearer now:

https://github.com/Ensembl/ensembl-vep/blob/master/modules/Bio/EnsEMBL/VEP/Parser/HGVS.pm#L214

PR #314, opened 2019-02-19
Automatically shift insertions/deletions within repeated regions to their most 3' position

aparton:
Lots of edge cases still to test, but this build is stable and ready to test


modules/Bio/EnsEMBL/Variation/BaseTranscriptVariation.pm
```diff
@@ -133,12 +133,13 @@ sub feature {
 =cut
 
 sub cdna_start {
-    my ($self, $cdna_start) = @_;
+    my ($self, $cdna_start, $toshift) = @_;
+    my $shifting_offset = defined($toshift) ? $toshift : 0;
```
ens-lgil: Could we simplify the code with the following line instead ?
```
$toshift ||= 0;
```
It seems to occur 6 times in the code

PR #316, opened 2019-02-20
Extend HGVSp - support protein deletions 

dglemos:


sarahhunt:
This works nicely, but there is now a lot of similar code in _parse_hgvs_protein_position and _parse_hgvs_protein_position_del. Can the reference look-up be extracted as a separate method that both of these can call? It would be useful for delins's too.

dglemos:
I'm going to look into this now.

modules/Bio/EnsEMBL/Variation/DBSQL/VariationFeatureAdaptor.pm
```diff
@@ -2270,6 +2250,90 @@ sub _parse_hgvs_protein_position{
...
+    
+    # get the last altered amino acid and position  
+    ($from2, $pos2, $to) = $second_residue =~ /^(\w+?)(\d+)(\w+?|\*)$/;
+  }
+  
+  # only one amino acid is deleted eg. Lys1110del
+  else{
+    ($from, $pos, $to) = $description =~ /^(\w+?)(\d+)(\w+?|\*)$/; 
+  } 
+  # conver three letter AA to single for both amino acids (convert first altered amino acid)
```
sarahhunt: $from isn't used, so does not need converting here.

PR #317, opened 2019-02-21
Feature/gerp for insertion

at7:
It makes sense to return GERP score of the location where insertion takes place.


at7:
merged with master and release/96

PR #318, opened 2019-02-21
Update annotationAdaptor.t

at7:
update to correct score


at7:
Merged to master and release/96

PR #324, opened 2019-03-04
Fix script for generating default vep examples for WEB VEP form

ima23:
- return example rsIDs of non failed variation
- remove pileup from formats generated by default, format not supported any more
- fix a typo/bug

### DIFF
--- a/scripts/misc/generate_vep_examples.pl
+++ b/scripts/misc/generate_vep_examples.pl
@@ -110,8 +110,11 @@ SPECIES: foreach my $species(@all_species) {
     my $name;
     my $sth = $real_vfa->db->dbc->prepare(qq{
       SELECT variation_name
-      FROM variation_feature
+      FROM variation_feature vf
+        LEFT JOIN failed_variation fv
+        ON vf.variation_id = fv.variation_id
       WHERE consequence_types LIKE ?
+      AND fv.variation_id IS NULL
       LIMIT 1
     });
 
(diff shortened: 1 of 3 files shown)


ens-lgil:
Actually we might want to add the `spdi` format as input

scripts/misc/generate_vep_examples.pl
```diff
@@ -110,8 +110,11 @@ =head1 CONTACT
     my $name;
     my $sth = $real_vfa->db->dbc->prepare(qq{
       SELECT variation_name
-      FROM variation_feature
+      FROM variation_feature vf
+        LEFT JOIN failed_variation fv
+        ON vf.variation_id = fv.variation_id
       WHERE consequence_types LIKE ?
+      AND fv.variation_id IS NULL
       LIMIT 1
```
helensch: Should this query be deterministic? If yes should include an order by on variation_name
https://dev.mysql.com/doc/refman/5.6/en/limit-optimization.html

ima23: Thanks for checking @helensch. Currently I get the same examples if the same database version is picked. I don't think there is a need of enforcing the sql to be deterministic. The generated examples are not particularly interesting besides being one a missense, one an intron variant and one a frameshift.

ima23:
Thanks for checking @ens-lgil. This PR is a bug fix for an existing jira ticket. I'd prefer to keep them separate but totally agree that if/ when`spdi` is going in the web vep forms, then this script will need an update.

> Actually we might want to add the `spdi` format as input

scripts/misc/generate_vep_examples.pl
helensch: If you want the same examples with a LIMIT you should use an order by. Getting the same examples at the moment is probably due to the index on name - just something to watch for in future if using LIMIT

ima23: Thanks @helensch, that's a good point! Will keep a note about it.

ens-lgil:
Should we wait for the addition of the SPDI examples before merging this PR ?

PR #325, opened 2019-03-08
deal with undeg BigWig parser

at7:
The new BigWig files are not in place yet but config files are already pointing to the new files and causing an error message on staging. This PR will fix the variant summary page.


at7:
I noticed that my branch is way behind. I will resubmit.

PR #326, opened 2019-03-08
deal with undefined BigWig parser

at7:
The new BigWig files are not in place yet but config files are already pointing to the new files and causing an error message on staging. This PR will fix the variant summary page.


ens-lgil:

PR #327, opened 2019-03-11
add conservation files

at7:
Adding all conservation score files for our variation species for release/96


ens-lgil:
Does it mean that we will have to update it every release ?
It's pointing at the `release-96` FTP directory.

at7:
Yes. But the compara team will very likely increase the number of species included in each alignment every release. Using the release version will allow for backwards compatibility, using the correct alignment for the given release. In the future the vcf_config file will be created automatically.

ens-lgil:
An other remark: should we use a different `id` for each entry ?
So far they are all the same.
e.g.: 
```
"id": "88_mammals.gerp_conservation_score",
"species": "bos_taurus",
```
Could become:
```
"id": "88_mammals.gerp_conservation_score_cow",
"species": "bos_taurus",
```

at7:
I don't think this is necessary. It will already be filtered by the species key.

ens-lgil:
Ok fine!

ens-lgil:

PR #328, opened 2019-03-11
fix a few cases in import

ima23:
import clinvar script fix for issue ENSVAR-1637:
- fix QC allele check id 14
- update log messaging
- add special character checking
- add hgvs parsing for _delIns_ of format  7:g.5978687_5978689delinsC and _del_ of format 9:g.137233961_137234061del

Sequence.pm:
- add coverage for 2 more hgvs parsing cases 
- add tests for all hgvs parsed cases

### DIFF
--- a/scripts/import/import_clinvar_xml
+++ b/scripts/import/import_clinvar_xml
@@ -839,18 +839,6 @@ sub enter_var{
     );
   $variation_adaptor->store($var);
 
-
-  foreach my $allele ( $ref_allele, $alt_allele){
-    $allele =~s/\s+//;
-    my $al = Bio::EnsEMBL::Variation::Allele->new
-      ( -variation_id   => $var->dbID(),
-        -allele         => $allele,
-        -adaptor        => $allele_adaptor,
-      );
-    $allele_adaptor->store($al);
-  }
-
-
   my $allele_str = $ref_allele ."/". $alt_allele;
 
   ## get slice for new variationfeature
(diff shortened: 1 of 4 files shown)


scripts/import/import_clinvar_xml
```diff
@@ -124,6 +124,23 @@ if (defined $done_file){
   }
 }
 
+my %special_characters = (
```
ens-lgil: Glad that this is simplier than we feared!

scripts/import/import_clinvar_xml
```diff
@@ -589,6 +613,15 @@ sub import{
   if(defined $record->{feature_info}->{dbSNP} && !defined $structvar){
     foreach my $rs (@{ $record->{feature_info}->{dbSNP} }){
 
+      # for haplotypes (multiple rsIDs) each rsID has it's own hgvs_g string
+      if (defined $record->{feature_info}->{'rs'.$rs}){
```
ens-lgil: We could use a global variable `my $rsid = 'rs'.$rs;` and/or assign:
```
my $record_feature = $record->{feature_info};
my $record_feature_info = $record_feature->{$rsid};

$record_feature->{hgvs_g} = $record_feature_info->{hgvs_g};
...
```

ima23: Thanks for the suggestion. Added it in the next commit.

PR #330, opened 2019-03-18
Change to FastaSequence.pm to allow fetching softmasked sequence from file

fergalmartin:
Updated code to allow a flag to preserve any softmasking in the fasta file when fetching sequence. This is to avoid querying the repeat_feature table via the API as asking for masked sequence incurs a join between repeat_feature and repeat_consensus that becomes expensive when there are many jobs running in parallel. The flag is passed through into the overriden ->seq into the _raw_seq subroutine. The code in _raw_seq has been slightly refactored to to the uc call after fetching the sequence, unless the preserve_masking flag has been passed. Example usage would be: my  = ->(undef,undef,undef,1)

### DIFF
--- a/modules/Bio/EnsEMBL/Variation/Utils/FastaSequence.pm
+++ b/modules/Bio/EnsEMBL/Variation/Utils/FastaSequence.pm
@@ -358,7 +358,7 @@ sub _get_fasta_db {
 # sequence is cached in 1MB range around the current slice
 sub _new_slice_seq {
   return sub {
-    my ($self, $start, $end, $strand) = @_;
+    my ($self, $start, $end, $strand, $preserve_masking) = @_;
     my ($seq, $length) = ('', 0);
 
     $start = $start ? ($self->start + $start) - 1 : $self->start;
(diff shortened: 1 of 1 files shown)


ens-lgil:
@fergalmartin, would it be possible to add a basic test to cover your changes in https://github.com/Ensembl/ensembl-variation/blob/master/modules/t/fastaSequence.t before the branching to 97 ?
Thanks.

fergalmartin:
@ima23, @ens-lgil I have asked @ens-carlos to look into putting in a test by Monday.

ens-carlos:
Would it be possible to modify the existing $Bin\/testdata/vep-cache/homo_sapiens/78_GRCh38/test.fa.gz file to lower case some of its bases on a region not used in any of the current tests or to have a softmasked version of any chromosome so I can point to it in the corresponding tests?

ima23:
@ens-carlos could you please add a small masked test fasta file in https://github.com/Ensembl/ensembl-variation/tree/master/modules/t/testdata? It's best to keep the vep-cache test data as it is. 
Thanks.

ens-carlos:
Yes, no problem. Thank you.

ens-carlos:
I've created a separate pull request to add the test.

PR #331, opened 2019-03-18
Feature/gerp root path

at7:
Set gerp_root_dir so we can only need to update the part of the file location which changes every release. Discussed with Anne [here](https://github.com/Ensembl/public-plugins/pull/194).


at7:
merged to master and release/96

PR #332, opened 2019-03-19
HGVS frameshift - throw useful message 

dglemos:


aparton:
I'm a little concerned that the error message might suggest that VEP doesn't support frameshifts at all, rather than solely when supplying an HGVSp input. Could we change it to something like:

HGVS Protein Input - Frameshifts are not supported


Other than that, I'm happy to merge.

PR #337, opened 2019-03-21
Add API support for AlleleSynonym

helensch:


modules/Bio/EnsEMBL/Variation/Variation.pm
```diff
@@ -1403,4 +1403,26 @@ sub update_attributes{
...
+  Status     : stable
+=cut
+sub get_all_AlleleSynonyms {
+    my $self = shift;
+
+    #Assert the adaptor reference
+    assert_ref($self->adaptor(),'Bio::EnsEMBL::Variation::DBSQL::BaseAdaptor');
+
+    # Get the AlleleSynonyms from the database
+    return $self->adaptor->db->get_AlleleSynonymAdaptor()->fetch_all_by_Variation($self);
```
at7: You need to check if $self->adaptor is defined. We do that in a few places across the Variation module.

at7: Let's address this in a post release fix if needed.

PR #339, opened 2019-03-21
Add the extra Core patches to the test Core database

ens-lgil:


ima23:
Tests fail due to recent changes to epigenome data structure.

PR #348, opened 2019-03-27
fix clinvar default phenotype

ima23:


ens-lgil:
Merged into `master` and `release/97`

PR #350, opened 2019-04-01
add test data and test new methods in TranscriptVariationAllele module

at7:
Finally adding tests for newish methods.


at7:

PR #352, opened 2019-04-04
Make the link to the GitHub module more generic (depends on the Ensem…

ens-lgil:
…bl release)


ens-lgil:
The variable is set in the webcode

at7:

PR #353, opened 2019-04-05
Skip variation feature having an empty string for at least one of the…

ens-lgil:
… alleles, e.g. /-

Work on my sandbox: http://ves-hx2-76.ebi.ac.uk:5060/Homo_sapiens/Variation/Mappings?v=rs1553414498

PR #354, opened 2019-04-05
Skip variation feature having an empty string for at least one of the…

ens-lgil:
… alleles, e.g. /-

Same as #353 for e97


at7:
Do we need this for 97? I thought it will be fixed with the next import.

ens-lgil:
I don't know ... however this issue might happen again in the future

PR #357, opened 2019-04-11
Import citations - small fix to print variants not found 

dglemos:


scripts/import/import_EPMC.pl
```diff
@@ -137,9 +137,10 @@ sub import_citations{
     my $data           = shift;
     my $type           = shift;
 
-    ## store variants not from this species
     open my $error_log, ">>$species\_$type\_notfound_" . log_time() . "\.log"|| die "Failed to open log file: $!\n";
-
+    ## store variants not from this species
+    open my $not_found, ">>rsID\_$type\_notfound\.csv"|| die "Failed to open file: $!\n";
```
sarahhunt: Without the species in the file name, the variants not found in one species will overwrite the variants not found in another species if the jobs are run in the same directory. Running 2 species at once becomes complex.
 
Without the timestamp, there is also a risk of over-writing.

scripts/import/import_EPMC.pl
```diff
@@ -137,9 +137,10 @@ sub import_citations{
...
+    print $not_found "rsID,PMCID,PMID\n";
```
sarahhunt: This header only matches EMPC data - UCSC uses a tab delimited format

scripts/import/import_EPMC.pl
helensch: Not using brackets and the lower precedence of || means that the file open is not checked

scripts/import/import_EPMC.pl
```diff
@@ -137,9 +137,16 @@ sub import_citations{
     my $data           = shift;
     my $type           = shift;
 
+    open (my $error_log, ">>$species\_$type\_notfound\_" . log_time() . "\.log") or die "Failed to open log file: $!\n";
```
sarahhunt: It would be easier to understand the file purpose if we replace 'notfound' with 'errors' in the file name

PR #359, opened 2019-04-12
Simplify the SQL queries used to remove the COSMIC data

ens-lgil:


scripts/import/remove_cosmic.pl
```diff
@@ -132,7 +127,7 @@ =head1 CONTACT
     AND     v.source_id = $source_id
```
sarahhunt: Are they ever in any set other than the cosmic set? This is probably a slow one, but deleting by set would be quick.

ens-lgil: Yes, I made quick simple changes. On e96 it took 1h with the old code

PR #360, opened 2019-04-12
Feature/ancestral alleles pipeline

at7:
A new e-hive pipeline for importing ancestral alleles!

### DIFF
--- a/modules/Bio/EnsEMBL/Variation/Pipeline/AncestralAlleles/Assign.pm
+++ b/modules/Bio/EnsEMBL/Variation/Pipeline/AncestralAlleles/Assign.pm
@@ -1,0 +1,168 @@
+=head1 LICENSE
+Copyright [1999-2015] Wellcome Trust Sanger Institute and the EMBL-European Bioinformatics Institute
+Copyright [2016-2019] EMBL-European Bioinformatics Institute
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+     http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+=cut
+
+
+=head1 CONTACT
+ Please email comments or questions to the public Ensembl
+ developers list at <http://lists.ensembl.org/mailman/listinfo/dev>.
+ Questions may also be sent to the Ensembl help desk at
+ <http://www.ensembl.org/Help/Contact>.
+=cut
+
+package Bio::EnsEMBL::Variation::Pipeline::AncestralAlleles::Assign;
+
+use strict;
+use warnings;
+
+use FileHandle;
+use Bio::EnsEMBL::Registry;
+use File::Path qw(make_path);
+use Data::Dumper;
+use Bio::DB::Fasta;
+use base qw(Bio::EnsEMBL::Variation::Pipeline::BaseVariationProcess);
+
+sub run {
+  my $self = shift;
+  my $species_name = $self->param('species_name');
+  my $batch = $self->param('batch');
+  my $id = $batch->{id};
+  my $start = $batch->{start};
+  my $end = $batch->{end};
+  my $ancestral_file = $self->param('ancestral_file');
+
+  $self->setup();
+  $self->dump_variation_features();
+  $self->assign_ancestral_alleles();
+  $self->cleanup();
+}
+
+sub setup {
+  my $self = shift;
+  my $batch = $self->param('batch');
+  my $id = $batch->{id};
+  my $compara_dir = $self->param('compara_dir'); 
+  my $species_dir = $self->param('species_dir');
+  my $ancestral_file_targz = $self->param('ancestral_file');
+  my $ancestral_file_notargz = $ancestral_file_targz;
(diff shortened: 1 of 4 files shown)


modules/Bio/EnsEMBL/Variation/Pipeline/AncestralAlleles/Init.pm
```diff
@@ -0,0 +1,199 @@
...
+  foreach my $vdba (@$vdbas) {
+      my $species_name = $vdba->species();
+      my $cdba =  $registry->get_DBAdaptor($species_name, 'core');
+      if ($cdba) {
+        $self->param('species', $species_name);
+        my $assembly = $self->get_assembly;
+        my ($ancestral_file) = grep { $_ =~ /$species_name/ && $_ =~ /$assembly/ } @ancestral_files;
+        if ($ancestral_file) {
+          $run_pipeline->{$species_name} = { file => $ancestral_file, assembly => $assembly};
+        }
```
sarahhunt: An else{} here to report if there is no ancestral file despite there being a core db could be useful.

modules/Bio/EnsEMBL/Variation/Pipeline/AncestralAlleles/Assign.pm
```diff
@@ -0,0 +1,168 @@
...
+    $self->param('species', $species_name);
+    my $vdba = $self->get_species_adaptor('variation');
+    my $dbc = $vdba->dbc;
+    my $sth = $dbc->prepare(qq{
+      SELECT vf.variation_feature_id, sr.name, vf.seq_region_start, vf.seq_region_end, vf.seq_region_strand, vf.ancestral_allele
+      FROM variation_feature vf, seq_region sr
+      WHERE vf.seq_region_id = sr.seq_region_id
+      AND vf.variation_feature_id >= ?
+      AND vf.variation_feature_id <= ?
+      ORDER BY vf.variation_feature_id;}, {mysql_use_result => 1}
```
sarahhunt: Is the order by quick enough for how useful it is?

modules/Bio/EnsEMBL/Variation/Pipeline/AncestralAlleles/Assign.pm
```diff
@@ -0,0 +1,168 @@
...
+    ) or die $dbc->errstr;
+    $sth->execute($start, $end) or die $dbc->errstr;
+    my ($vf_id, $seq_region, $start, $end, $strand, $aa);
+    $sth->bind_columns(\($vf_id, $seq_region, $start, $end, $strand, $aa));
+    while ($sth->fetch) {
+      $aa ||= 'NULL';
+      print $fh join("\t", ($vf_id, $seq_region, $start, $end, $strand, $aa)), "\n";
```
sarahhunt: It should be quicker to calculate the AA without writing the input to disk and reading it again.

modules/Bio/EnsEMBL/Variation/Pipeline/AncestralAlleles/Assign.pm
```diff
@@ -0,0 +1,168 @@
...
+        if ($aa && $aa =~ m/^[ACGTacgt]+$/) {
+          $aa = uc $aa;
+          $set_aa_to_null = 0;
+          print $fh "UPDATE variation_feature SET ancestral_allele='$aa' WHERE variation_feature_id=$vf_id;\n" if ($aa ne $old_aa);
+        } else {
+          print $err "No AA in fasta $vf_id\n";
+        }
+      }
+    }
+    if ($set_aa_to_null && $old_aa ne 'NULL') {
```
sarahhunt: It would be useful to check that the entire batch is not being set to null.

sarahhunt:
@at7 - Looks good! I've merged this into e!97 and master and added some suggestions for speed and additional input data checking.

modules/Bio/EnsEMBL/Variation/Pipeline/AncestralAlleles/Init.pm
at7: We can also add an entry to meta if the species has ancestral allele data. This should make it more reliable to get all species with ancestral data.

modules/Bio/EnsEMBL/Variation/Pipeline/AncestralAlleles/Assign.pm
at7: This is not needed anymore now that we annotate per variation feature. It was relevant for the old script. I will remove that. But it is not a bottleneck.

modules/Bio/EnsEMBL/Variation/Pipeline/AncestralAlleles/Assign.pm
at7: This is a precautionary measure in case we have to deal with any failures and rerun the job. Then we don't need to connect to the database again but can annotate from file. But I agree that it is a manageable size to annotate on the fly.

modules/Bio/EnsEMBL/Variation/Pipeline/AncestralAlleles/Assign.pm
at7: Do you think this would mean something is wrong with the ancestral FASTA file or that particular job has problems reading from disk? Usually it will flag up in the final count comparison if the number of NULLs has gone up. But it is true that it will be difficult tell if there was a systematic error.

sarahhunt: I'm thinking about particular jobs having problems reading from disk. I know we shouldn't see this.

PR #368, opened 2019-05-02
Add import uniprot synonyms 

dglemos:


scripts/import/import_variant_synonyms
```diff
@@ -147,6 +197,356 @@ sub download_PharmGKB{
...
+      my $gene = $1;
+      my $uniprot_id = $2;
+      my $rs_id = $4;
+      my $phenotype = $5;
+      
+      # If no rsId was given, will attempt to get one by looking up the Uniprot id in the synonym table
+      if ($rs_id ne '-') {
+        push(@{$synonym{$rs_id}},$uniprot_id);
+      }
+      #else {
```
ens-lgil: Maybe you could removed these commented lines as we don't import phenotype data for this source anymore

scripts/import/import_variant_synonyms
```diff
@@ -147,6 +197,356 @@ sub download_PharmGKB{
...
+
+=cut 
+
+sub download_UniProt{
+
+  my $http     = HTTP::Tiny->new();
+  my $response = $http->get('https://www.uniprot.org/docs/humsavar.txt');
+  die "Failed to pick up file\n" unless $response->{success};
+
+  open my $out, ">humsavar.txt" ||die "Failed to write data locally : $!\n";
```
helensch: Not using brackets and the lower precedence of || means that the file open is not checked

scripts/import/import_variant_synonyms
```diff
@@ -147,6 +197,356 @@ sub download_PharmGKB{
...
+  #my %result = ('synonyms' => \%synonym, 'phenotypes' => \@phenotypes);
+  my %result = ('synonyms' => \%synonym);
+  return \%result;
+}
+
+sub get_dbIDs {
+  my $rs_ids = shift;
+  my $db_adaptor = shift;
+  
+  my $id_stmt = qq{
```
helensch: order is non-deterministic when using LIMIT without order. Does it matter?

scripts/import/import_variant_synonyms
dglemos: Thanks! I deleted it.

scripts/import/import_variant_synonyms
dglemos: Thanks for spotting this (again).

scripts/import/import_variant_synonyms
dglemos: It doesn't.

scripts/import/import_variant_synonyms
```diff
@@ -147,20 +197,337 @@ sub download_PharmGKB{
...
+      $source_id,
+      ?
+    )
+  };
+
+  my $stmt_update = qq{
+    UPDATE
+      variation_feature 
+    SET 
+      variation_set_id = CONCAT(variation_set_id,',$variation_set_id')
```
helensch: CONCAT will return NULL if variation_set_id IS NULL, currently it not allowed to be NULL but this may change in future see ENSVAR-1091

scripts/import/import_variant_synonyms
```diff
@@ -147,20 +197,337 @@ sub download_PharmGKB{
...
+  my $stmt_update = qq{
+    UPDATE
+      variation_feature 
+    SET 
+      variation_set_id = CONCAT(variation_set_id,',$variation_set_id')  
+    WHERE
+      variation_id=?
+  }; 
+
+  print "SQL: $stmt_update\n";
```
helensch: Is print stament for logging or debugging?

dglemos: It's printed for debugging but I think there's no need to keep it.

scripts/import/import_variant_synonyms
```diff
@@ -112,13 +111,64 @@ elsif($source_name eq 'dbSNP'){
...
+
+  my @ids; 
+  my %synonym;
+  my @phenotypes; 
+  if (exists($result->{'synonyms'})) {
+    %synonym = %{$result->{'synonyms'}};
+    # To get all the ids of the source (Uniprot)
+    @ids = keys(%synonym);
+  }
+  # not used right now
```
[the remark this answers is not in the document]
dglemos: If we don't use it, I'll remove them.

scripts/import/import_variant_synonyms
```diff
@@ -147,20 +197,337 @@ sub download_PharmGKB{
...
+
+    $db_adaptor->dbc->do($ins_stmt); 
+    $variation_set_id = $db_adaptor->dbc->db_handle->{'mysql_insertid'};
+ 
+  } 
+  
+  return $variation_set_id; 
+}
+
+sub add_synonyms_uniprot {
```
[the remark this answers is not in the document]
dglemos: I changed add_synonyms_uniprot to use add_synonym (used by import_synonyms)

scripts/import/import_variant_synonyms
dglemos: I'm going to write a note in ENSVAR-1091 about this script and leave it for now like this.

scripts/import/import_variant_synonyms
```diff
@@ -147,20 +197,337 @@ sub download_PharmGKB{
...
+      SELECT
+        attrib_id
+      FROM
+        attrib
+      WHERE
+        value = '$set_name'
+      LIMIT 1
+    };
+    my $sth_attrib = $db_adaptor->dbc->prepare($stmt_attrib);
+    $sth_attrib->execute();
```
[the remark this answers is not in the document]
dglemos: I missed this! Also, I changed the source_name from 'ph_uniprot' to 'UniProt variants'. When searching for source_name in attrib table it will probably fail because attrib.value = 'ph_uniprot'.

scripts/import/import_variant_synonyms
```diff
@@ -112,39 +111,364 @@ elsif($source_name eq 'dbSNP'){
...
+    $id_sth->fetch();
+    
+    # If we couldn't find the rs_id, look in the synonym table
+    if (!defined($var_id)) {
+      $syn_sth->bind_param(1,$rs_id,SQL_VARCHAR);
+      $syn_sth->execute();
+      $syn_sth->bind_columns(\$var_id,\$var_name);
+      $syn_sth->fetch();
+    }
+    # warn "$rs_id - no mapping found in variation db\n" unless $var_id ;
```
helensch: Should this debug comment be removed?

scripts/import/import_variant_synonyms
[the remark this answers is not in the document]
helensch: I suggest a check is still needed to check that an attrib_id has been found, else it will still insert 0 if not found

scripts/import/import_variant_synonyms
dglemos: It should be there, I was testing and forgot to put it again.

scripts/import/import_variant_synonyms
helensch: Can the update to variation_set_id be done with either a
CONCAT(ifnull(variation_set_id,''), ',$variation_set_id')
or
CONCAT_WS(',', variation_set_id, '$variation_set_id')
instead of a check and then an insert which may have race conditions

scripts/import/import_variant_synonyms
```diff
@@ -112,39 +111,364 @@ elsif($source_name eq 'dbSNP'){
...
+
+    $mapping{$rs_id} = [$var_id,$var_name] if $var_id && $var_name;
+  }
+  
+  return \%mapping;
+} 
+
+=head2 add_set_uniprot
+
+add uniprot variation set
```
helensch: Suggest changing uniprot to UniProt in description to match other sub descriptions

scripts/import/import_variant_synonyms
dglemos: Thanks for the suggestion!

sarahhunt:
This looks good. As discussed, it would be better to default source version to the run date and warn that this is happening unless a version is supplied on the command line.

PR #369, opened 2019-05-03
Fix to check variation feature before fetching allele string 

dglemos:


ens-lgil:
You might need to send a PR against `postreleasefix/97` as well

at7:
The hack is not merged into release/97. I hope the data has been fixed and we don't need the check anymore.

PR #371, opened 2019-05-08
Fix/dumps

at7:
Fixing a few bugs before starting 97 dumps.
https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-1672
https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-1983


ens-lgil:
Merged into `release/97` and `master`

PR #374, opened 2019-05-08
Fix RefSeq HGVS - get xref transcripts

dglemos:


modules/Bio/EnsEMBL/Variation/DBSQL/VariationFeatureAdaptor.pm
```diff
@@ -1773,6 +1773,11 @@ sub fetch_by_hgvs_notation {
     $reference =~ s/\.\d+//g;
    }
 
+  # Old code only worked with right transcript version
+  # Delete transcript version to avoid problems
+  my $reference_no_version = $reference;
+  $reference_no_version =~ s/\.[0-9]+//;
```
sarahhunt: It would be better to keep the version in the sequence look up if possible, so we return exacly what was requested.

modules/Bio/EnsEMBL/Variation/DBSQL/VariationFeatureAdaptor.pm
```diff
@@ -1861,12 +1871,21 @@ sub fetch_by_hgvs_notation {
...
+    # Fetch xref transcript
+    if(!defined($transcript) && $reference =~ /NP/){
+      @transcripts = @{$transcript_adaptor->fetch_all_by_external_name($reference_no_version)};
+    } 
+
     # support some malformed HGVS
     if(!defined($transcript)) {
       # Seeing transcripts erroneously submitted with p. changes
-      if($reference =~ /ENST|NM/){
+      if($reference =~ /ENST/){
```
sarahhunt: This needs to be backwards compatible as most species have not been reloaded in the new format - we need fetch_by_stable_id() for dog, for example, and fetch_all_by_external_name for human and maybe mouse.

PR #375, opened 2019-05-09
Feature/check tabix exists

at7:
This has been applied to release/96 https://github.com/Ensembl/ensembl-variation/pull/349 and is still missing from master/release/97.


ima23:
merged to `release/97` and `master`

PR #384, opened 2019-05-15
Fix HGVS refseq transcripts

dglemos:
This is a fix for e!96 
#385 is for e!97


sarahhunt:
Thanks @dglemos - can you add a test?

ens-lgil:
@sarahhunt, is it ok to merge it ?

modules/Bio/EnsEMBL/Variation/TranscriptVariationAllele.pm
```diff
@@ -1023,7 +1023,10 @@ sub hgvs_protein {
   ## checks complete - start building term
 
   ### get reference sequence and add seq version unless LRG
-  $hgvs_notation->{ref_name} = $tr->translation->display_id();
+  my $tr_id = $tr->translation->display_id();
+  $tr_id =~ s/.*NP_/NP_/;
```
sarahhunt: This won't fix all records, see cds-NP_001185471.1-2.1:p.Asn46Ser here :

http://rest.ensembl.org/variant_recoder/human/rs764041571?content-type=application/json

Does something like this work:

https://github.com/Ensembl/ensembl-vep/blob/release/96/modules/Bio/EnsEMBL/VEP/AnnotationSource/Database/Transcript.pm#L668

dglemos: I thought the lazy fix would work. It's fixed now.

PR #385, opened 2019-05-15
Fix HGVS refseq transcripts (97) 

dglemos:


modules/Bio/EnsEMBL/Variation/TranscriptVariationAllele.pm
```diff
@@ -1023,9 +1023,13 @@ sub hgvs_protein {
   ## checks complete - start building term
 
   ### get reference sequence and add seq version unless LRG
-  my $tr_id = $tr->translation->display_id();
-  $tr_id =~ s/.*NP_/NP_/; 
-  $hgvs_notation->{ref_name} = $tr_id;
+  $hgvs_notation->{ref_name} = $tr->translation->display_id();
+  
+  ### get RefSeq identifiers
```
sarahhunt: If we already have an Ensembl protein accession, we don't want to spend time looking up xrefs too.

We could check if the transcript stable_id ~/ ENST/LRG/ and use the display id if it does or look for the xref if it does not.

PR #386, opened 2019-05-16
Expand ref allele

dglemos:


ens-lgil:
I believe this is the same as #378

PR #387, opened 2019-05-21
Updates to PhenotypeFeatureAdaptor.pm

aparton:


modules/Bio/EnsEMBL/Variation/DBSQL/PhenotypeFeatureAdaptor.pm
```diff
@@ -1074,6 +1074,92 @@ sub fetch_all {
   
 }
 
+sub get_PhenotypeFeatures_by_location {
+  my $self = shift;
+  my $seq_region_id = shift;
+  my $seq_region_start = shift;
+  my $seq_region_end = shift;
+
+  throw("Cannot fetch attributes without dbID") unless defined($seq_region_id);
```
ens-lgil: Shouldn't it be 
```
... unless (defined(seq_region_id) && defined(seq_region_start) && defined(seq_region_end));
```
?

modules/Bio/EnsEMBL/Variation/DBSQL/PhenotypeFeatureAdaptor.pm
```diff
@@ -1074,6 +1074,92 @@ sub fetch_all {
...
+  throw("Cannot fetch attributes without dbID") unless defined($seq_region_start);
+  throw("Cannot fetch attributes without dbID") unless defined($seq_region_end);
+
+  my $attribs = {};
+
+  my $extra_sql = " pf.seq_region_id = '$seq_region_id' AND pf.seq_region_start >= '$seq_region_start' AND pf.seq_region_end <= '$seq_region_end' ";
```
ens-lgil: Shouldn't be necessary to use the single quotes for numeric values

modules/Bio/EnsEMBL/Variation/DBSQL/PhenotypeFeatureAdaptor.pm
```diff
@@ -1074,6 +1074,92 @@ sub fetch_all {
...
+  return $self->generic_fetch("$extra_sql");
+}
+
+sub get_PhenotypeFeatureAttribs_by_location {
+  my $self = shift;
+  my $seq_region_id = shift;
+  my $seq_region_start = shift;
+  my $seq_region_end = shift;
+
+  throw("Cannot fetch attributes without dbID") unless defined($seq_region_id);
```
ens-lgil: Same remark as above

modules/Bio/EnsEMBL/Variation/DBSQL/PhenotypeFeatureAdaptor.pm
ens-lgil: Could we replace "Cannot fetch attributes without dbID" by something like "Cannot fetch attributes without location information" ?

modules/Bio/EnsEMBL/Variation/DBSQL/PhenotypeFeatureAdaptor.pm
```diff
@@ -1074,6 +1074,88 @@ sub fetch_all {
   
 }
 
+sub get_PhenotypeFeatures_by_location {
```
ens-lgil: Minor request that I overlooked in the first review: could we have some perldoc for this method ?
This is would also allow to have the method documented in Doxygen.

modules/Bio/EnsEMBL/Variation/DBSQL/PhenotypeFeatureAdaptor.pm
```diff
@@ -1074,6 +1074,88 @@ sub fetch_all {
...
+  my $attribs = {};
+
+  my $extra_sql = " pf.seq_region_id = $seq_region_id AND pf.seq_region_start >= $seq_region_start AND pf.seq_region_end <= $seq_region_end ";
+
+  $extra_sql = $self->_is_significant_constraint($extra_sql);
+
+  return $self->generic_fetch("$extra_sql");
+}
+
+sub get_PhenotypeFeatureAttribs_by_location {
```
ens-lgil: Same comment about adding Perldoc to this method

modules/Bio/EnsEMBL/Variation/DBSQL/PhenotypeFeatureAdaptor.pm
```diff
@@ -1074,6 +1074,88 @@ sub fetch_all {
...
+  my $self = shift;
+  my $seq_region_id = shift;
+  my $seq_region_start = shift;
+  my $seq_region_end = shift;
+
+  throw("Cannot fetch attributes without seq region information") unless defined($seq_region_id) && defined($seq_region_start) && defined($seq_region_end);
+
+  my $extra_sql = $self->_is_significant_constraint();
+
+  my $sth = $self->dbc->prepare(qq{
```
sarahhunt: It would save time to restrict by source_id in phenotype_feature.

modules/Bio/EnsEMBL/Variation/DBSQL/PhenotypeFeatureAdaptor.pm
```diff
@@ -1074,6 +1074,114 @@ sub fetch_all {
...
 }
 
+=head2 get_PhenotypeFeatures_by_location
+
+  Arg [1]    : int $seq_region_id
+  Arg [2]    : int $seq_region_start
+  Arg [3]    : int $seq_region_end
+  Description: Retrieves PhenotypeFeatures by genomic location
+               
+  Returntype : hashref of array of types => phenotype_feature count
```
sarahhunt: This appears to be returning phenotype features rather than counts.

modules/Bio/EnsEMBL/Variation/DBSQL/PhenotypeFeatureAdaptor.pm
```diff
@@ -1074,6 +1074,114 @@ sub fetch_all {
...
+  Exceptions : none
+  Caller     : general
+
+=cut
+
+sub get_PhenotypeFeatures_by_location {
```
sarahhunt: This method looks very similar to fetch_all_by_Slice_type - is it needed?

sarahhunt: This method is not used in the corresponding PR.

modules/Bio/EnsEMBL/Variation/DBSQL/PhenotypeFeatureAdaptor.pm
```diff
@@ -1074,6 +1074,116 @@ sub fetch_all {
...
+
+  my $extra_sql = " pf.seq_region_id = $seq_region_id AND pf.seq_region_start >= $seq_region_start AND pf.seq_region_end <= $seq_region_end ";
+  $extra_sql .= " AND pf.source_id = $source_id " if defined($source_id);
+
+  $extra_sql = $self->_is_significant_constraint($extra_sql);
+
+  return $self->generic_fetch("$extra_sql");
+}
+
+=head2 get_PhenotypeFeatureAttribs_by_location
```
sarahhunt: The name and docs on this method are confusing as it doesn't return all PhenotypeFeature attribs by location - it currently restricts to those with an associated allele. Names and docs should reflect function. 

We only use capitalisation like 'PhenotypeFeatureAttribs' when we are returning  PhenotypeFeatureAttrib objects, so it is not appropriate here as we don't have PhenotypeFeatureAttrib objects.

It would be clearer and more efficient to have a method  'get_clinsig_alleles_by_location' returning only alleles and their associated clinical significances per location. Then state in the docs that is has a specific caller.

PR #398, opened 2019-06-18
patch test dbs for release/98

at7:


at7:
Merged with master and release/98

PR #399, opened 2019-06-19
Updates to Equivalent Allele Pipeline

helensch:
Failed variations are now included.
Duplicates are reported - caused by resetting failed jobs
In future review if a unique index on variation_attrib is needed


at7:
Merged with master and release/98

PR #401, opened 2019-06-21
Removes start/stop_retained consequences from COSMIC variants

aparton:


ens-lgil:
Should we create a test for the non `ATGC-` alleles ?

ens-lgil:

PR #404, opened 2019-06-26
updates clinvar import

ima23:
- extract version based on date within xml
- read from .xml or .xml.gz
- fix case where ref== alt, no variation will be created
- replace xml reader with XML::LibXML::Reader 
- handle special character encoding parsing and replacement (added new Bio/EnsEMBL/Variation/Utils/SpecialChar.pm + specialChar.t)


ima23:
Will close this PR to `postreleasefix/98` and make one to master. 

Part of this PR is replicated in PR https://github.com/Ensembl/ensembl-variation/pull/406 to `postreleasefix/98` which is the code I used for import.

PR #411, opened 2019-07-10
Change variant recoder error message - spdi invalid sequence id

dglemos:


modules/Bio/EnsEMBL/Variation/DBSQL/VariationFeatureAdaptor.pm
```diff
@@ -2477,6 +2477,7 @@ sub fetch_by_spdi_notation{
   if($count_separator > 3){ throw ("Could not parse the SPDI notation $spdi. Too many elements present"); } 
   elsif($count_separator < 3){ throw ("Could not parse the SPDI notation $spdi. Too few elements present"); } 
 
+  my $seq_id = $sequence_id;
```
ens-lgil: Maybe a bit picky but the variable name `seq_id` is shorter than `sequence_id` and basically it's the way around into these 2 variables.
I would suggest a slightly more self explanatory variable name for `$seq_id`, e.g. something like `$whole_sequence_id` or `$full_sequence_id` or even `$raw_sequence_id` (to mimic a bit what we have in ensembl-io `get_raw_seqname()`).

dglemos: I'll change that. Thanks.

aparton:
Closing & reopening to set off the travis run again - I believe this should pass now

PR #414, opened 2019-07-15
reduce hive capacity

at7:
I'm reducing the hive capacity which reduces the number of db connections. I noticed problems when running TV for opossum with hive_capacity of 150. Reducing hive_capacity to 50 solved the connection problems.


modules/Bio/EnsEMBL/Variation/Pipeline/VariationConsequence_conf.pm
```diff
@@ -263,7 +263,7 @@ sub pipeline_analyses {
           },
           { -logic_name => 'by_gene_transcript_effect',
             -module => 'Bio::EnsEMBL::Variation::Pipeline::TranscriptEffect',
-            -hive_capacity  => 150,
+            -hive_capacity  => 50,
```
helensch: If the hive_capcity is being reduced to 50, should the analysis_capacity also be reduced from 60 to 50?

ens-lgil:

PR #415, opened 2019-07-17
extend special char list - code run e98

ima23:
- general phenotypes special char list extended
- OMIA url updated


ens-lgil:

PR #416, opened 2019-07-18
Bugfix98/ld

at7:
https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-2324


modules/Bio/EnsEMBL/Variation/DBSQL/LDFeatureContainerAdaptor.pm
```diff
@@ -349,6 +349,11 @@ sub fetch_by_VariationFeatures {
     $self->{_pairwise}->{$vf->seq_region_start} = 1;
     $self->{_pairwise_vf_name}->{$vf->variation_name} = 1;
   }  
+
+  # if pairwise store the first vf name to retain order in output
+  if (scalar @$vfs == 2) {
+    $self->{_pairswise_first_vf_name} = $vfs->[0]->variation_name;
```
helensch: Might be easier to read later if _pairswise was _pairwise (without the s) as for other keys.

PR #417, opened 2019-07-22
Transcript Variation COSMIC fix

aparton:


helensch:
Looks good.  
This was for ENSVAR-2318. A PR will be done for e98. Could a test be added for allele_string of COSMIC that flagged be issue?

PR #419, opened 2019-07-28
most_severe_OverlapConsequence set alphabetically for equal ranks

helensch:
For consequences with the same highest rank, the first alphabetically is returned
This a fix for JIRA ENSVAR-2119 (98) and ENSVAR-909 (99)
Should a PR be made on postreleasefix/98?


helensch:
Replaced by #434

PR #421, opened 2019-07-29
Task98/post pipeline updates

at7:
also don't compute motif variation overlaps for somatic variants because adding cosmic allele strings (COSMIC) into the motif is causing errors because of non supported nucleotide chars.


at7:

PR #422, opened 2019-07-29
Bugfix98/gnomad neg frequencies

at7:
we ended up with multiple mappings in our gnomad files. The API counts all the frequencies for all observed alleles which results in negative allele frequencies in some cases:
https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-2117
-- for now ignore each new row with duplicated alt alleles
(-- need to look into how to clean up gnomad files)


modules/Bio/EnsEMBL/Variation/VCFCollection.pm
```diff
@@ -632,13 +632,19 @@ sub get_all_Alleles_by_VariationFeature {
...
-
-    foreach my $pop(@pops) {
+     POP : {
+     foreach my $pop(@pops) {
 
       # this allows for an empty population name to be looked up as e.g. AF
       my $raw_name = exists($pop->{_raw_name}) ? $pop->{_raw_name} : $pop->name;
       my $suffix = $raw_name ? '_'.$raw_name : '';
       my $pop_id = $pop->dbID;
+      foreach my $allele (values %allele_map) {
```
sarahhunt: A comment would be useful here to explain this handles multiple lines with the same alt by only considering the first

PR #423, opened 2019-07-30
Task98/software upgrade

at7:


aparton:
Happy with this, I'll wait until the installer is ready before merging however.

PR #425, opened 2019-07-31
Updates to allow test to be run with database intact

helensch:
Fix for ENSVAR-2338
Also update to only use a single vf when testing store as only single vf returned by fetch_by_dbID


helensch:
Reducing the test code (not the number of tests) has reduced the coverage.

PR #426, opened 2019-07-31
recognize http too

at7:
Web noticed a problem with retrieving plant data from http://ftp.sra.ebi.ac.uk... I checked with them and adding http to the regex solves the problem. This will go live as soon it is merged. I will create a PR against 98 branches too.


at7:
Tests pass on postrelasefix/98. Errors are related to version mismatch in travis config.

PR #427, opened 2019-07-31
recognize http too

at7:
Same as https://github.com/Ensembl/ensembl-variation/pull/426


ens-lgil:

PR #428, opened 2019-08-01
Bugfix98/data dumps

at7:
A few minor updates


aparton:
Happy with the code changes. Only suggestion is that we should link to the current_variation folder  in the comments rather than the r97 specific one. i.e. ftp://ftp.ensembl.org/pub/current_variation/gvf/homo_sapiens/

at7:
I need to keep the release number in the path because this is the last release which contains the HapMap data dumps.

at7:
merged with master and release/98

PR #430, opened 2019-08-01
Fix source dbSNP - different versions 

dglemos:
There's one source 'dbSNP151' in production db that was being created in human db when importing synonyms - also, there's no human synonym data for that source in production.  
Changed code to only check for sources with synonym data.

### DIFF
--- a/scripts/import/import_variant_synonyms
+++ b/scripts/import/import_variant_synonyms
@@ -109,11 +109,12 @@ elsif($source_name eq 'dbSNP') {
   my %data; # synonym_name => {rs_name} & {source_id} 
 
   # get data from human_synonym table (in production db) 
+  # some synonyms name are duplicated - from different sources 
   foreach my $result (@{$synonyms_output}){  
-    $data{$result->[1]}{rs_name} = $result->[0]; 
-    $data{$result->[1]}{source_id} = $result->[2]; 
+    $data{$result->[0].' '.$result->[1]}{rs_name} = $result->[0]; 
+    $data{$result->[0].' '.$result->[1]}{source_id} = $result->[2]; 
   }
- 
+
   # get all available sources from production database  
   my $sources = get_all_sources($dba, $prod_dbc);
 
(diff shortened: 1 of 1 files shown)


scripts/import/import_variant_synonyms
```diff
@@ -103,7 +103,8 @@ elsif($source_name eq 'dbSNP') {
       '-driver'  =>  'mysql' 
     ); 
 
-  my $synonyms_output = query_synonyms_data($prod_dbc, 'rs_name,synonym_name,source_id', 'human_synonym'); 
+  my $sql_query = "SELECT rs_name,synonym_name,source_id FROM human_synonym"; 
+  my $synonyms_output = query_synonyms_data($prod_dbc, $sql_query);
```
sarahhunt: The data structure made from $synonyms_output assumes synonym_name is unique, but it is not.

scripts/import/import_variant_synonyms
```diff
@@ -103,16 +103,18 @@ elsif($source_name eq 'dbSNP') {
...
+  my $synonyms_output = query_synonyms_data($prod_dbc, $sql_query); 
 
   my %data; # synonym_name => {rs_name} & {source_id} 
 
   # get data from human_synonym table (in production db) 
+  # some synonyms name are duplicated - from different sources 
   foreach my $result (@{$synonyms_output}){  
-    $data{$result->[1]}{rs_name} = $result->[0]; 
-    $data{$result->[1]}{source_id} = $result->[2]; 
+    $data{$result->[0].' '.$result->[1]}{rs_name} = $result->[0];
```
sarahhunt: Is this a unique set - do no sources use the same name for the same variant?

sarahhunt:
The UniProt import is looking up variant ids twice. 

Can you remove  get_dbIDs and at ~line 451 switch :
-  foreach my $rs_id (keys %{$variation_ids}) {
+  foreach my $rs_id (keys %{$synonyms}) {

scripts/import/import_variant_synonyms
dglemos: $data{$result->[1]} is not unique because the same synonym name has different rsid - it was being overwritten.  
$data{$result->[0].' '.$result->[1]} is unique.

sarahhunt: This would be dangerous if we were adding further synonyms of this type to the production db, but as it is a static list it should not be a problem.

PR #432, opened 2019-08-06
Fix issue when the reference allele is missing in the variant allele …

ens-lgil:
…strings

Similar to the PR #372


ima23:
Merge into `master` and `release/98`

PR #434, opened 2019-08-09
most_severeOverlapConsequence set alphabetically for equal ranks

helensch:
For consequences with the same highest rank, the first alphabetically is returned.
This is a fix for JIRA ENSVAR-2119 (98) and ENSVAR-909 (99)
PR is on postreleasefix/98 and replaces #419


aparton:
Happy with what's there, but we may need a little more. We can discuss it when we're all in the office.

Happy with this as a fix for the most_severe_OverlapConsequence methods, however we have other rank sorting code in Variation/Utils/VEP. Similar code exists in the OutputFactory. Potentially the stats as well.

at7:
I will merge this PR for 98. But for another time could BaseVariationFeatureOverlapAllele::get_all_OverlapConsequences return a sorted list already?

at7:
merged with master and release/98

PR #437, opened 2019-08-14
Check that alt_cds is not a sequence of Ns, preventing HGVSp calculation

aparton:


modules/Bio/EnsEMBL/Variation/TranscriptVariationAllele.pm
```diff
@@ -1698,6 +1698,10 @@ sub _get_fs_peptides {
   ### get CDS with alt variant
   my $alt_cds = $self->_get_alternate_cds();
   return undef unless defined($alt_cds);
+  unless ($alt_cds->seq =~ /([ACGT-])+/) {
```
ens-lgil: Should it be `$alt_cds->seq =~ /^([ACGT-])+$/)` to make sure there are only ATGC- characters in the CDS sequence ?

aparton: I don't think so, with a sequence like NNNNANNNN or AAAANAAAA then we want it to carry on. This is just a crude 'if everything is Ns, stop' check.

at7:

PR #438, opened 2019-08-15
Remove some hard-coded configuration (moved to the VEP_plugins/plugin…

ens-lgil:
…_config.txt file)


at7:

PR #441, opened 2019-08-19
update dbnsfp key used by protein function pipeline

at7:
When I was running the CADD and dbNSFP protein function pipeline on a fresh human database for release 98 I noticed that one of the attribs was still using the old name dbnsfp_meta_svm instead of dbnsfp_meta_lr.


at7:

PR #444, opened 2019-08-27
dbSNP import fixes

helensch:
Fixed for issues importing b152:
- do not import duplicated variation_features
- for 1000Genomes set minor allele to "-" when empty
- for 1000Genomes flip minor allele

### DIFF
--- a/scripts/import/dbSNP_v2/load_dbsnp.pl
+++ b/scripts/import/dbSNP_v2/load_dbsnp.pl
@@ -895,7 +895,7 @@ sub import_variation {
   # Determine the values for maf, minor_allele, minor_allele_count
   my ($maf, $minor_allele, $minor_allele_count);
   if (defined $freq) {
-      $minor_allele = $freq->{'minor_allele'};
+      $minor_allele = $freq->{'minor_allele'} || '-';
       $maf = format_frequency($freq->{'MAF'});
       $minor_allele_count = $freq->{'minor_allele_count'};
   }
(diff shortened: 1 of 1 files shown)


scripts/import/dbSNP_v2/load_dbsnp.pl
```diff
@@ -225,6 +276,81 @@ sub parse_refsnp {
...
+    print $fh join("\t", $var_name, 'INFO',
+      'review_status: ' . $align_diff->{'review_status'}), "\n";
+    return;
+  }
+
+  my $grch37_loc = $align_diff->{'grch37_loc'};
+  my $grch38_loc = $align_diff->{'grch38_loc'};
+  if ($grch37_loc != $grch38_loc) {
+    print $fh join("\t", $var_name, 'SKIP',
+      "Num of GRCh37 loc ($grch37_loc) differ num of GRCh38 loc ($grch38_loc)"), "\n";
```
sarahhunt: Can this message be made clearer? Does this mean if there is 1 mapping on GRCh37 and 2 on GRCh38, no allele is returned but if there are 2 on both it is?

helensch: Yes means that if the number of mappings differ between GRCh37 and GRCh38 the allele not be  reverse_complemented, no allele returned, minor allele will stay as orginal
If the number of mapping the same between GRCh37 and GRCh38 the allele will be reverse complemented -even if number of mappings both 2. Should this be restricted to 1 mapping and 1 opposite alignment for GRCh37 and GRCh38?

helensch:
Replaced by #451

PR #445, opened 2019-08-30
fix vcf position when alleles contain common first base 

sarahhunt:


at7:

PR #446, opened 2019-09-03
Bugfix98/ld windowsize

at7:
A few bug fixes which prevented passing on the specified window size to the LD script. The problem was revealed by the ld variation endpoint which was cutting off the window size unexpectedly. I checked that the REST LD tests pass with the updates.


aparton:
Merged with release/98 and master

PR #462, opened 2019-09-30
Fix bugs in script import citations 

dglemos:


scripts/import/import_EPMC.pl
```diff
@@ -744,7 +744,7 @@ sub clean_publications{
 
     # If there is publications without title, delete them 
     if($n_title_null != 0){
-      my $title_null_sth = $dba->dbc->prepare(qq[ select publication_id from publication where title is null or title = '' ]);
+      my $title_null_sth = $dba->dbc->prepare(qq[ select publication_id, pmid from publication where title is null or title = '' ]);
```
ima23: Where is the added pmid used?

scripts/import/import_EPMC.pl
```diff
@@ -770,12 +781,10 @@ sub remove_publications{
 
     $var_id_sth->execute()||die;
     my $get_variation_ids = $var_id_sth->fetchall_arrayref();
-    my $variation_ids = $get_variation_ids->[0]->[0];
+    my $variation_ids = $get_variation_ids->[0];
```
ima23: Could a comment be added?

scripts/import/import_EPMC.pl
dglemos: When removes the publication it write in a file the publication_id and in this case the tile is null so can't write the title, write the pmid instead. Writing only the publication_id is not info enough.

scripts/import/import_EPMC.pl
dglemos: Done. Thanks.

ima23:

PR #464, opened 2019-10-03
fix gwas variation sets

ima23:


ens-lgil:
Merged into `master` and `release/99`

PR #466, opened 2019-10-16
add salmon PRJEB34225 from EVA

at7:
The config update is needed for API users. I will have a separate PR for public-plugins.


at7:

PR #467, opened 2019-10-16
Task99/ld for remote vcf

at7:
-- EVA VCF files for salmon use non ensembl chromosome names. Enabled use of synonyms for LD calculation
-- We need to change into tmp dir for calculation of LD from remote VCF files


at7:
Example for salmon on my sandbox http://ves-hx2-76.ebi.ac.uk:5040/Salmo_salar/Location/LD/ajax?db=core;r=ssa01%3A1338365-1346896;pop1=4;pop2=1

modules/Bio/EnsEMBL/Variation/DBSQL/LDFeatureContainerAdaptor.pm
```diff
@@ -423,15 +421,29 @@ sub _fetch_by_Slice_VCF {
...
         push @regions, $loc_string;
       }
       my $files_arg = join(',', @files); 
       my $regions_arg = join(',', @regions);
       my $number_of_files = scalar @files;
       my $window_size = $self->max_snp_distance;
-      $cmd = "$bin -f $files_arg -r $regions_arg -s $number_of_files -l $sample_string -w $window_size";
 
+      my $working_dir= cwd();
+      chdir $self->temp_path if ($self->temp_path);
```
helensch: suggest checking the return value from chdir

modules/Bio/EnsEMBL/Variation/DBSQL/LDFeatureContainerAdaptor.pm
```diff
@@ -423,15 +421,29 @@ sub _fetch_by_Slice_VCF {
...
-        my $loc_string = sprintf("%s:%i-%i", $slice->seq_region_name, $slice->start, $slice->end);
+        my $chr = $slice->seq_region_name;
+        if ($vc->use_seq_region_synonyms) {
+          my @chr_in_vcf_file = @{$vc->_vcf_parser_obj($vcf_file)->{tabix_file}->seqnames};
+          my @synonyms = ();
+          foreach my $synonym (@{$vc->_get_synonyms_by_chr($chr)}) {
+            push @synonyms, $synonym if (grep {$_ eq $synonym} @chr_in_vcf_file);
+          }
+          $chr = $synonyms[0];
+          warn "use_seq_region_synonyms is set. Found more than one synonym for sequence name $chr" if (scalar @synonyms > 1);
```
helensch: Suggest putting checks of @synonyms in an if/elsif loop. If @synonyms > 1, do no need to check if no synonyms.

helensch:
Not part of the code changes, but was noted during debugging.
There is a $DB::single = 1;
https://github.com/Ensembl/ensembl-variation/blob/b522f2eb4e557c8841fbd74952a2f94bf6adf355/modules/Bio/EnsEMBL/Variation/DBSQL/LDFeatureContainerAdaptor.pm#L330
Suggest that this is removed.

modules/t/LDFeatureContainerAdaptor.t
```diff
@@ -338,4 +338,25 @@ foreach my $result (@$results_fetch_by_Slice) {
...
+-chromosomes => [9],
+-sample_prefix => '1000GENOMES:phase_1:',
+-adaptor => $vca
+);
+
+$vca->remove_VCFCollection_by_ID('ld_without_rs_in_vcf');
+$vca->add_VCFCollection($c);
+delete $ldfca->{_cached};
+$ldfc = $ldfca->fetch_by_Slice($slice, $population);
+$ld_values = $ldfc->get_all_ld_values;
```
helensch: Is this needed? $ld_values not used in following code.

modules/Bio/EnsEMBL/Variation/DBSQL/LDFeatureContainerAdaptor.pm
at7: I didn't use if/elsif because I don't have anything for the else clause to report. I'm also testing for no synonyms because use_seq_region_synonyms is set which means that if there wasn't a synonym found for one of the chromosomes in the VCF file it would be worth finding out why we don't have a synonym for it.

helensch: An else would not be needed, code would continue. Was just noting if  @synonyms > 1, then no need to check scalar @synonyms == 0. 
if (scalar @synonyms > 1) {
  warn "use_seq_region_synonyms is set. Found more than one synonym for sequence name $chr";
} elsif (! @synonyms) {
  warn "use_seq_region_synonyms is set. But didn't find synonym for sequence name $chr";
}

at7: That makes sense. I will update the PR with your suggestion.

modules/Bio/EnsEMBL/Variation/DBSQL/LDFeatureContainerAdaptor.pm
```diff
@@ -423,15 +419,33 @@ sub _fetch_by_Slice_VCF {
...
       }
       my $files_arg = join(',', @files); 
       my $regions_arg = join(',', @regions);
       my $number_of_files = scalar @files;
       my $window_size = $self->max_snp_distance;
-      $cmd = "$bin -f $files_arg -r $regions_arg -s $number_of_files -l $sample_string -w $window_size";
 
+      my $working_dir = cwd();
+      my $return_chdir = chdir $self->temp_path if ($self->temp_path);
+      warn("ERROR: Couldn't change working directory (" . $self->temp_path . "): $!\n") if (!$return_chdir);
```
helensch: The warn is being reported with no chdir as $return_chdir is only set if ($self->temp_path).
The warn only needs to be given if the chdir was attempted and failed.

```
if ($self->temp_path) {
  my $return_dir = chdir $self->temp_path;
  if (! $return_dir) {
     warn("ERROR: Couldn't change working directory (" . $self->temp_path . "): $!\n")
  }
}
```

```

File: modules/Bio/EnsEMBL/Variation/DBSQL/LDFeatureContainerAdaptor.pm [hunk 5]
```diff
@@ -508,6 +522,9 @@ sub _fetch_by_Slice_VCF {
       # glob without closing it.
       close LD;
 
+      $return_chdir = chdir $working_dir if ($self->temp_path);
+      warn("ERROR: Couldn't change working directory ($working_dir): $!\n") if (!$return_chdir);
```
helensch: As for other chdir, warn should only be given if chdir done

modules/Bio/EnsEMBL/Variation/DBSQL/LDFeatureContainerAdaptor.pm
```diff
@@ -153,7 +152,6 @@ sub vcf_executable {
 sub temp_path {
```
helensch: Suggest adding some documentation for sub temp_path including some of what in PR description.
When is $TMP_PATH set? 
Would other calls be expecting to be /tmp

helensch:
Thanks for the documentation updates. Using example on the sandbox getting "No such file or directory" error.

at7:
I was on a different branch and switched back to task99/LD_for_remote_vcf now on my sandbox

helensch:
Thanks for changing the branch. An error is reported when doing pairwise LD from 2 variants selected from LD plot.
ssa01:1340651:G_A:PRJEB34225
ssa01:1341931:G_T:PRJEB34225
Could not fetch variant object for ssa01:1341931:G_T:PRJEB34225

at7:
This is a separate issue. We cannot fetch by VF U think. I will crate a jira ticket.

helensch:
The podchecker on LDFeatureContainerAdaptor.pm is returning an error

at7:

PR #470, opened 2019-10-22
Task99/update gerp files

at7:
-update gerp file version
-update zebrafinch assembly
-add gerp file for salmon


ima23:

PR #472, opened 2019-10-24
Fix species option to be string

helensch:
ld_tool species option failing as species being treated as flag. From example
./ld_tool --show_populations --species human


at7:

PR #473, opened 2019-10-25
Fix HGVSp when using 'NM' transcript 

dglemos:
HGVSp is failing with transcript NM because transcript has to be fetched by external name (example NM_000088.3:p.Val1374Ala) 
http://rest.ensembl.org/variant_recoder/human/NM_000088.3:p.Val1374Ala?content-type=application/json 

Test: http://ebi-cli-003:3000/variant_recoder/human/NM_000088.3:p.Val1374Ala?content-type=application/json


ima23:
Can there be a test added for this? Thanks.

ima23:

PR #474, opened 2019-10-29
fix region parser

at7:


tools/linkage_disequilibrium/ld_tool
```diff
@@ -240,6 +240,7 @@ sub region {
   my @populations = @{$config->{_populations}};
   foreach my $region (@regions) {
     my ($chromosome, $start, $end) = split /\s+/, $region;
+    die "Invalid format for region. Expected chromosome:start-end." unless ($chromosome && $start && $end);
```
helensch: Suggest that check start <= end

PR #476, opened 2019-10-29
Bugfix98/check existing missing alleles

at7:
Deals with bug reported by Andrey for  VEP tool [Missing reference or alternate sequence]
Input like "chr3    146187575       146187575       A/0     +       het" was causing a problem for compare_existing call.

### DIFF
--- a/modules/Bio/EnsEMBL/Variation/Utils/Sequence.pm
+++ b/modules/Bio/EnsEMBL/Variation/Utils/Sequence.pm
@@ -949,7 +949,7 @@ sub align_seqs {
 sub trim_sequences {
   my ($ref, $alt, $start, $end, $empty_to_dash, $end_first) = @_;
 
-  throw("Missing reference or alternate sequence") unless $ref && $alt;
+  throw("Missing reference or alternate sequence") unless defined $ref && defined $alt;
 
   $start ||= 0;
   $end ||= $start + (length($ref) - 1);
--- a/modules/t/sequence_utils.t
+++ b/modules/t/sequence_utils.t
@@ -109,6 +109,12 @@ is_deeply(
   'trim_sequences - empty_to_dash'
 );
 
+is_deeply(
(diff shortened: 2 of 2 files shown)


modules/t/sequence_utils.t
```diff
@@ -109,6 +109,12 @@ is_deeply(
   'trim_sequences - empty_to_dash'
 );
 
+is_deeply(
+  trim_sequences(qw(A 0)),
+  ['A', 0, 0, 0, 0],
+  'trim_sequences Except 0 as alt allele'
```
helensch: Possible typo - suggest accept rather than except

at7: Thanks. Fixed now

modules/Bio/EnsEMBL/Variation/Utils/Sequence.pm
```diff
@@ -949,7 +949,7 @@ sub align_seqs {
 sub trim_sequences {
   my ($ref, $alt, $start, $end, $empty_to_dash, $end_first) = @_;
 
-  throw("Missing reference or alternate sequence") unless $ref && $alt;
+  throw("Missing reference or alternate sequence") unless defined $ref && defined $alt;
```
helensch: Is  an empty alt also now accepted?
trim_sequences('A','')

at7: I don't think it matters for the trim_sequences call.

helensch: Before the code change, the trim_sequences('A','') was throwing a warning. Now it will not. If does not matter then ok.

at7: It doesn't only throw a warning but an error and stops the whole VEP analysis. I don't think this method is the right place for allele checks.

PR #478, opened 2019-10-29
Bugfix99/check existing missing alleles

at7:
same as #476


at7:

PR #479, opened 2019-10-30
Fix/schema attrib

helensch:
Update attrib for release 99


at7:

PR #480, opened 2019-10-30
use 99 branch

at7:


at7:
If we update the branch to release/99 in .travis.yml for the postreleasefix/99 branch it will overwrite what it is in the master .travis.yml file when we merge postreleasefix/99 into master.

PR #482, opened 2019-10-31
add missing chr for salmon

at7:
Thanks @helensch for spotting the missing chromosomes


at7:

PR #484, opened 2019-11-05
cosmic import updates - e100

ima23:
- adapt code for latest format
- inversion start=end+1
- add variation sets

### DIFF
--- a/scripts/import/import_cosmic.pl
+++ b/scripts/import/import_cosmic.pl
@@ -228,7 +228,7 @@ sub get_equivalent_class {
   my $start = shift;
   my $end   = shift;
 
-  my @type_parts = split(' ',$type);
+  my @type_parts = split(',',$type);
   $type = $type_parts[0];
 
   my $class = $default_class;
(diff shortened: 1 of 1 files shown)


scripts/import/import_cosmic.pl
```diff
@@ -217,7 +228,7 @@ sub get_equivalent_class {
   my $start = shift;
   my $end   = shift;
 
-  my @type_parts = split(' ',$type);
+  my @type_parts = split(',|\s',$type);
   $type = $type_parts[0];
```
dglemos: Some COSMIC variants have more than one class splited by ',' in this case only the first class is used. Is this the expected behavior? One example: `"Complex - compound substitution,Substitution - intronic"` after import class = sequence_alteration

scripts/import/import_cosmic.pl
```diff
@@ -310,11 +321,13 @@ sub add_phenotype {
 }
 
 sub get_variation_set_id {
-  
+  my $type = shift;
+
   # Check if the COSMIC set already exists, else it create the entry
```
dglemos: Comment says set is created if it's not found but set is not created.

scripts/import/import_cosmic.pl
ima23: Yes, this is expected.

PR #486, opened 2019-11-14
fix transcript ablation -e98

ima23:
Fix for https://github.com/Ensembl/ensembl-vep/issues/627. Transcript_ablation sometimes annotated as 'stop_lost,coding_sequence_variant,5_prime_UTR_variant,3_prime_UTR_variant,intron_variant'.

### DIFF
--- a/modules/Bio/EnsEMBL/Variation/BaseVariationFeatureOverlapAllele.pm
+++ b/modules/Bio/EnsEMBL/Variation/BaseVariationFeatureOverlapAllele.pm
@@ -438,14 +438,18 @@ sub _bvfo_preds {
   my ($vf_start, $vf_end) = ($bvf->{start}, $bvf->{end});  
   my ($min_vf, $max_vf) = $vf_start > $vf_end ? ($vf_end, $vf_start) : ($vf_start, $vf_end);
 
-  if ( $bvf->isa('Bio::EnsEMBL::Variation::StructuralVariationFeature') ){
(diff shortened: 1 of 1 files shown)


aparton:
Is there a reason why we return after we calculate that the variant completely overlaps the feature? This change doesn't add 'within_feature' to the predicates list, which may have unforeseen consequences further down the line

ima23:
> Is there a reason why we return after we calculate that the variant completely overlaps the feature? This change doesn't add 'within_feature' to the predicates list, which may have unforeseen consequences further down the line

Thanks for checking the PR. The change does add 'within_feature' unless there is a return, that was the issue. The overlapping features were considered same as 'within_feature'.  Do you have an example for unforeseen consequences further down the line?

aparton:
My presumption was that we would want the _bvfo_preds method to describe the variant as both 'complete_overlap' and 'within_feature', as there are a collection of annotations within the output factory that are only returned if a variant is described as 'within_feature'. Is this correct, or is the removal of 'within_feature' the fix rather than the addition of 'complete_overlap'?

This changes how different SVs are annotated as well. For example:
_perl vep -id '3 10164900 10243745 DUP' --database_ 
no longer provides a cDNA position in the output.

And I know I'm being picky with this (as I think it's caused by a rounding error), but I get an OverlapPC of 100% and the list of consequences with the input command:
 perl vep -id '3 10164900 10243744 DEL' --database

ima23:
Wormbase example: `chrIII 10193562 10201986 DEL` and `T20G5.7.1` transcript.

PR #488, opened 2019-11-15
fix transcript ablation -e99

ima23:
same as https://github.com/Ensembl/ensembl-variation/pull/486


aparton:
Merged into release/99 and master

ima23:
@aparton : can you please confirm if this was indeed merged to master? I don't see it in `release/100` nor in `master`. Thanks.

PR #489, opened 2019-11-19
Feature/indel assignment

helensch:
For alleles from SPDI in dbSNP JSON imports some indels are more precisely insertions or deletions

PR for ENSVAR-2042


sarahhunt:
I would expect TA/T to be a deletion.

Ideally, TAC/TC and AAA/AAA/AA/A would also be deletions, but we can assume we wouldn't normally see strings like this.

helensch:
Currently   'AGCG/-/A'              => 'sequence_alteration',
and based on this   'CTT/C'      => 'indel',  # Not a deletion as first base kept
that is why TA/T  and TAC/TC not assigned as deletion

AAA/AAA is classified as a substitution
AA/AA an SNV
Currently alleles the same as reference are excluded from assignment to deletions, insertions, indels

aparton:
Looks good. The tests cover all of the single alternate allele cases that I'd expect, so VEP shouldn't be affected.

PR #492, opened 2019-11-20
Use assembly in dbSNP imports

helensch:


scripts/import/dbSNP_v2/load_dbsnp.pl
```diff
@@ -92,9 +92,11 @@
 my ($num_lines) = parse_dbSNP_file($config);
 
 # Close open filehandles
-for my $ma_type ('update', 'log') {
-  my $fh = $config->{join('-', 'ma', $ma_type, 'fh')};
-  $fh->close();
+if ($config->{'assembly'} eq 'GRCh38') {
```
sarahhunt: A comment here would be helpful

scripts/import/dbSNP_v2/load_dbsnp.pl
```diff
@@ -129,19 +131,25 @@ sub init_reports {
     die("$config->{'error_file'} already exists. Please rename or delete\n");
   }
 
+  # For GRCh38 the 1000Genomes minor allele needs flipping
```
sarahhunt: This suggests they all need reverse complimenting

scripts/import/dbSNP_v2/load_dbsnp.pl
```diff
@@ -1587,6 +1616,16 @@ sub get_seq_region_names {
   return (\%seq_region);
 }
 
+# For a given chr look up the seq_region
+sub get_seq_region_chr {
+  my ($dbh, $chr) = @_;
```
sarahhunt: This relies on only one seq_region record for Y being imported from core to variation. We should probably check somewhere.

scripts/import/dbSNP_v2/load_dbsnp.pl
helensch: A comment has been added.

scripts/import/dbSNP_v2/load_dbsnp.pl
helensch: A comment has been added and a ticket created to check this at the start of the script.

scripts/import/dbSNP_v2/load_dbsnp.pl
helensch: A comment has been added

PR #493, opened 2019-11-21
Task99/pipeline updates

at7:
Updates to pipelines that were run for 99.


aparton:
Merged into release/99 and master

PR #495, opened 2019-11-29
Patch SV - add frequency columns

dglemos:


sql/table.sql
```diff
@@ -1192,6 +1192,8 @@ CREATE TABLE structural_variation_association (
 @column somatic                          Flags whether this structural variation is known to be somatic or not
 @column breakpoint_order                 Defines the order of the breakpoints when several events/mutation occurred for a structural variation (e.g. somatic mutations)
 @column length                           Length of the structural variant. Used for the variants with a class "insertion", when the size of the insertion is known.
+@column allele_freq                      The max allele frequency of the minor allele of this variant, as reported by gnomAD.
```
sarahhunt: The frequency is associated with this allele and this study - other studies apart from gnomAD may have data. A better comment would be something like 'The frequency reported for this allele in this study '

sql/table.sql
```diff
@@ -1192,6 +1192,8 @@ CREATE TABLE structural_variation_association (
...
+@column allele_count                     The number of samples the minor allele of this variant is found in.
```
sarahhunt: Similarly, a better comment would be 'The number of times this allele is observed in this study'

sql/table.sql
dglemos: Thanks. Changed description.

sql/table.sql
dglemos: Thanks, description changed.

dglemos: Thanks, description changed.

PR #496, opened 2019-12-02
Update structural variants import

dglemos:


scripts/import/import_all_sv_data.pl
```diff
@@ -343,19 +349,20 @@ sub study_table{
...
     $year_desc = $2 if (defined($2));
   }
-        
+
   my $author_desc;
   $author_desc = "$first_author $year_desc " if ($first_author and defined($year_desc));
-  $author = (split('_',$author))[0] if ($author !~ /.+_et_al_.+/);
-  
+  # Don't think this is needed
+  # $author = (split('_',$author))[0] if ($author !~ /.+_et_al_.+/);
```
at7: Which values do we expect for $author?

scripts/import/import_all_sv_data.pl
```diff
@@ -420,15 +431,15 @@ sub study_table{
...
     my $external_link_sql;
     if ($external_link eq 'NULL') {
-      if ($study_xref && $study_xref !~ /NULL/i && $study_xref ne '') {
+      # if ($study_xref && $study_xref !~ /NULL/i && $study_xref ne '') {
         $external_link_sql = '';
-      }
-      else {
-        $external_link_sql = "external_reference='$external_link',";
-      }
+      # }
```
at7: Is there a reason why we keep this as a comment?

scripts/import/import_all_sv_data.pl
```diff
@@ -420,15 +431,15 @@ sub study_table{
     elsif ($assembly eq $target_assembly && $study_desc =~ /^(.+)\s\[remapped/) {
       $study_desc = $1;
     }
-    
+
     my $external_link_sql;
     if ($external_link eq 'NULL') {
-      if ($study_xref && $study_xref !~ /NULL/i && $study_xref ne '') {
+      # if ($study_xref && $study_xref !~ /NULL/i && $study_xref ne '') {
```
at7: Do we need to keep this?

scripts/import/import_all_sv_data.pl
```diff
@@ -749,46 +760,51 @@ sub structural_variation_set {
...
-    
-    # Variation set "High quality"
+
+    my $sets_1kg = $dbVar->selectrow_arrayref(qq{
+         SELECT DISTINCT vssv.variation_set_id
+         FROM variation_set_structural_variation vssv, structural_variation sv, $temp_table t
+         WHERE vssv.structural_variation_id=sv.structural_variation_id AND sv.variation_name=t.id
+       });
+    foreach my $s_id (@$sets_1kg) {
+      push (@var_set_list,$s_id);
```
at7: The brackets are not needed: push @var_set_list, $s_id;

scripts/import/import_all_sv_data.pl
```diff
@@ -1847,7 +1898,7 @@ sub cleanup {
   }
   
   # Column "tmp_clinic_name" in structural_variation
-  my $sth2 = $dbVar->prepare(qq{ SELECT count(*) FROM $sv_table WHERE clinical_significance is NULL AND $tmp_sv_clin_col is not NULL});
+  my $sth2 = $dbVar->prepare(qq{ SELECT count(*) FROM $sv_table WHERE (clinical_significance is NULL or clinical_significance = '') AND $tmp_sv_clin_col is not NULL});
```
at7: or should be upper case to be consistent. => 'NULL OR'

scripts/import/import_all_sv_data.pl
```diff
@@ -1487,65 +1524,88 @@ sub parse_9th_col {
...
     $info->{is_somatic} = 1 if ($text =~ /var_origin=Somatic/i || !$text);
   }
-  
+
   return $info;
 }
 
-
+# Replace weird characters
+# dbVar weird characters are still in the files - replace them here
```
at7: Can that not also go into replace_char?

scripts/import/import_all_sv_data.pl
dglemos: It shouldn't be commented on. This line is needed to get the author's name for the description.

scripts/import/import_all_sv_data.pl
dglemos: I didn't think about putting there because they are very specific to dbVar files. I don't think they were found anywhere else but I'll put it there to be more consistent.

scripts/import/import_all_sv_data.pl
dglemos: It should. Thanks.

scripts/import/import_all_sv_data.pl
dglemos: This is the same situation as the previous comment. Shouldn't be commented, I forgot to fix it before PR.

scripts/import/import_all_sv_data.pl
dglemos: Same as previous.

scripts/import/import_all_sv_data.pl
```diff
@@ -159,7 +163,7 @@ =head1 CONTACT
 $cs_version_number =~ s/\D//g;
 
 # variation set
-my %var_set = ('pilot1' => 31, 'pilot2' => 32);
+# my %var_set = ('pilot1' => 31, 'pilot2' => 32);
```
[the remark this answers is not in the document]
dglemos: No, I'll delete it.

PR #500, opened 2019-12-05
Updated to cosmic import -e100

ima23:
- add evidence and display info at import
- insert default phenotype with 'tumour' attrib (patch 99_100c)


scripts/import/import_cosmic.pl
```diff
@@ -308,10 +311,27 @@ sub get_source_id {
   return $source_id[0];
 }
 
+sub get_attrib_id {
+  my ($type, $value) = @_;
```
dglemos: You could add a comment saying which values are accepted for $type and $value. For $type you use 'evidence' and 'phenotype_type' but are more values accepted?

ima23: The method can retrieve any `type, value` pairs based on the attrib and attrib_type tables. In this script I'm only using `evidence` and `phenotype_type` types.

PR #505, opened 2019-12-12
import clinvar updates for - e101

ima23:
- add evidence attribs to variation and variation feature
- add variation set ids to variation feature 
- add phenotype evidence const to Sequence utils
- update VariationAdaptor update method to avoid undef errors   
- created jira ticket for VariationFeatureAdaptor to include variation sets in the future ENSVAR-2717

### DIFF
--- a/scripts/import/import_clinvar_xml
+++ b/scripts/import/import_clinvar_xml
@@ -1050,6 +1050,7 @@ sub enter_var{
       $ref_allele = $refSlice->seq;
       $allele_str = $ref_allele ."/". $alt_allele;
   }
+  print "Saving allele string: $vf->allele_string()\n";
 
   my $var = Bio::EnsEMBL::Variation::Variation->new
     ( -name              => $dbSNP,
--- a/scripts/import/import_clinvar_xml
+++ b/scripts/import/import_clinvar_xml
@@ -110,6 +110,9 @@ clean_old_data() if defined $clean;
 my $omim_set_id   = get_set($dbh, 'ph_omim');
 my %omimVarSet;
 
+my $pheno_set_id   = get_set($dbh, "ph_variants"); #All phenotype/disease variants
+my $clinvar_set_id = get_set($dbh, "ClinVar"); #All ClinVar variants
+
 ## default ClinVar phenotype description if non available
 my $default_pheno  = "ClinVar: phenotype not specified";
 my $haplotype_type = "Haplotype";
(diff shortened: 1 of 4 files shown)


scripts/import/import_clinvar_xml
```diff
@@ -988,6 +1050,7 @@ sub enter_var{
       $ref_allele = $refSlice->seq;
       $allele_str = $ref_allele ."/". $alt_allele;
   }
+  print "Saving allele string: $vf->allele_string()\n";
```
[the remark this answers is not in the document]
ima23: No, it was a stashed lined from previous work. I removed it now.

scripts/import/import_clinvar_xml
```diff
@@ -274,7 +286,7 @@ sub get_set{
 
   if (defined $set_id->[0]->[0] ){
     ## remove old set content
-    $set_del_sth->execute($set_id->[0]->[0]) if defined $clean;
+    $set_del_sth->execute($set_id->[0]->[0]) if defined $clean && $set eq 'ph_omim';
```
sarahhunt: It's confusing to only delete the OMIM set here. Is it safe to continue to delete all the ClinVar sets here? If not, a comment would be useful.

ima23: Thanks for checking @sarahhunt. Have removed it in the next commit.

scripts/import/import_clinvar_xml
```diff
@@ -766,6 +806,28 @@ sub update_variation_set{
...
+      UPDATE variation_feature
+      SET
+        variation_set_id = CONCAT_WS(",", variation_set_id, ?),
+        evidence_attribs = CONCAT_WS(",", evidence_attribs, '$pheno_evidence_id')
+      WHERE variation_feature_id = ? ] );
+  my $new_sets = "$pheno_set_id,$clinvar_set_id";
+  if ($omimVarSet{$vf->variation->dbID}) {
+    $new_sets .=",$omim_set_id";
+  }
+  $vf_upd_sth->execute( $new_sets,  $vf->dbID );
```
sarahhunt: Which db server is this working on? I tried it on prod4 and got:
 DBD::mysql::st execute failed: Data truncated for column 'variation_set_id' at row 1 at scripts/import/import_clinvar_xml line 826.

It looks like '23,30,24' is being treated as a string.

ima23: Thanks for checking. There is a setup difference between prod4 and the other servers which leads to this error and not on the others, I created a jira ticked for it ENSVAR-2979.

PR #506, opened 2019-12-17
temp fix for pheno - e99

ima23:
Ignore phenotype is_significant flag, Jira: ENSVAR-2732


at7:
The phenotypeFeatureAdaptor.t test is failing.

PR #507, opened 2019-12-17
Fix extraction of SNV from VCF when adjacent to synonym for SNV

helensch:
This is the fix for release 99, the fix for release 98 was in PR #503


at7:

PR #510, opened 2019-12-19
replace deprecated code

at7:

PR #511, opened 2019-12-19
support * as base in allele string

at7:
The tests will fail. The API codes need to be updated in order for the test to pass. https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-2747


at7:
Fixed with https://github.com/Ensembl/ensembl-variation/pull/510

PR #512, opened 2019-12-19
Prevent repeated calculation of flanking sequence length

aparton:
And additional comments


dglemos:
Merged into `master` and `release/100`.

PR #513, opened 2019-12-19
check that allele string exists

at7:
SVs don't have an allele string and a warning was printed for uninitialised value in comparison in the start_retained_variant  predicate. I create the same PR for postreleasefix/100.


aparton:
The same line is written in the stop_retained method, the same check should probably be added there

PR #518, opened 2020-01-09
adding attached VCf for wheat

gnaamati:
Adding remote VCF files for wheat variation

### DIFF
--- a/modules/Bio/EnsEMBL/Variation/DBSQL/vcf_config.json
+++ b/modules/Bio/EnsEMBL/Variation/DBSQL/vcf_config.json
@@ -1,5 +1,29 @@
 {
   "collections": [
+     {
+        "id": "820K",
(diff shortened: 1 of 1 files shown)


gnaamati:
Hi Anja, I tested this against the variation API but will do so again (the link to the FTP changed). This also works on my sandbox.
I am trying to get the LD display to work for release 100, I assumed that the variation handover date was the same for PRs. If not I can put this in postreleasefix/100 or if this is an issue do this in 101, I'm not certain the LD will work for 100.
I will also remove strict_name_match and sample_prefix.

gnaamati:
This works against the variation API.

at7:
If this should go into 100 can you please do a PR against postreleasefix/100? Thanks

at7:
@gnaamati Could you please let us know if this PR is still needed?

PR #519, opened 2020-01-13
escape the special regex character

at7:
Bug fix for https://github.com/Ensembl/ensembl-vep/issues/671


ima23:

PR #522, opened 2020-01-15
Remove flanking seq table

ima23:

### DIFF
--- a/scripts/import/import_vcf.pl
+++ b/scripts/import/import_vcf.pl
@@ -1924,10 +1924,6 @@ sub variation {
       $var->{ancestral_allele} = $data->{info}->{AA} eq '.' ? undef : uc($data->{info}->{AA});
     }
 		
-		# add in some hacky stuff so flanking sequence gets written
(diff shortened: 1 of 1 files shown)


sarahhunt:
If you look at the history, EnsemblVariationSchema.txt and remove_wrong_variations.sql haven't been updated since 2007 and 2010. They are no longer used, so did not need updating. We need to trawl for obsolete code in coming months.

sarahhunt:
For import_vcf, 
- flank can be removed from the config at line 216
- the flanks option should be removed from the GetOptions list
- the upstream/downstream setting at 1932 isn't needed

ima23:
Thanks @sarahhunt for checking. I have removed the updated to the 2 files that have not been changed for a while. Current PR contains now only the  previous + mentioned changes to import_vcf.pl

sarahhunt:
All of the upstream/downstream flank setting code from comment 'add in some hacky stuff so flanking sequence gets written' can be removed

ima23:
Thanks @sarahhunt for checking. I've removed that old section in the next commit.

PR #524, opened 2020-01-16
Prevent shifting if no_shift is undefined

aparton:


at7:
If ( defined($args{'-no_shift'}) && !$args{'-no_shift'} ) is true does that mean we shift?

aparton:
If --no_shift is set to 1, then we shouldn't shift. If no shift is set to 0, then we should shift. 

The change should handle the undefined case. Previously, if it was undefined, this would trigger shifting code. We want to not shift by default in that case

at7:
We shouldn't pass a value to --no_shift. no_shift means no shift. --no_shift 0 doesn't make sense unfortunately. Is this something a VEP user needs to decide? Or is this an internal argument?

at7:
Some tests in variant_effect.t fail

ima23:

PR #526, opened 2020-01-20
Task100/remapping updates

at7:
Compare tables and table columns between old and new variation database. We had a problem remapping macaque from 97 to 100 because of schema patches for structural variation tables.

### DIFF
--- a/modules/Bio/EnsEMBL/Variation/Pipeline/Remapping/PreRunChecks.pm
+++ b/modules/Bio/EnsEMBL/Variation/Pipeline/Remapping/PreRunChecks.pm
@@ -68,6 +68,26 @@ sub run {
     . "AND sr.seq_region_id = sra.seq_region_id; ";
   $vdba->dbc()->sql_helper()->execute_update(-SQL => $sql, -PARAMS => [] );
 
+
+  my $table_columns_newasm = $self->get_table_columns($registry);
+  $registry->load_all($self->param('registry_file_oldasm'));
+  my $table_columns_oldasm = $self->get_table_columns($registry);
+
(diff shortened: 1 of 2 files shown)


modules/Bio/EnsEMBL/Variation/Pipeline/Remapping/PreRunChecks.pm
```diff
@@ -68,6 +68,26 @@ sub run {
...
+  # Check that tables and columns are the same between assemblies or if the old databases need to be patched to the new schema
+  foreach my $db_type (qw/variation core/) {
+    if (join(',', sort keys %{$table_columns_newasm->{$db_type}}) ne join(',', sort keys %{$table_columns_oldasm->{$db_type}})) {
+      die "Lists of $db_type tables differ between assemblies. You probably need to patch the  $db_type database.\n"
+    } else {
+      foreach my $table_name (keys %{$table_columns_newasm->{$db_type}}) {
+        if ($table_columns_newasm->{$db_type}->{$table_name} ne $table_columns_oldasm->{$db_type}->{$table_name}) {
+          my $columns_oldasm = $table_columns_oldasm->{$db_type}->{$table_name};
+          my $columns_newasm = $table_columns_newasm->{$db_type}->{$table_name};
+          die "Lists of columns (old $columns_oldasm, new $columns_newasm) differ between assemblies for table $table_name. You probably need to patch the $db_type database.\n";
```
dglemos: This means that if a table has different columns, first we need to patch the old db?

at7: Yes. I will update confluence. If that error occurs you need to copy the previous database from mirror to ens-var-prod, update the registry file and patch the old database before running again.

PR #527, opened 2020-01-21
add ESP GRCh37 VCF

at7:
Add ESP GRCh37 VCF location.
Background: We don't import ESP data into our variation database anymore. Get frequencies from VCF instead.


at7:
I'm going to add the update into task100/gnomad_v3

PR #528, opened 2020-01-22
Clear up empty consequences

aparton:
Shifting code has led to multiple occasions where consequence types within the transcript variation table can be empty. This tidies up many of these:

1) modules/Bio/EnsEMBL/Variation/TranscriptVariationAllele.pm - Prevents shifting for allele strings labelled INSERTION, DELETION, HGMD, COSMIC & CNV
2) Includes variants that start off outside of upstream and shifted into it as upstream
3) Doesn't print variants that return no consequences and are further than the given max_distance away from the transcript (i.e. variants that are shifted out of downstream)

This wont clear up all empty consequences. There are still issues with multi-allelic variants where the second alt allele is shifted into/out of protein_coding regions, but there's no quick fix for that one, so it's likely to be 101.

### DIFF
--- a/modules/Bio/EnsEMBL/Variation/Utils/VariationEffect.pm
+++ b/modules/Bio/EnsEMBL/Variation/Utils/VariationEffect.pm
@@ -347,6 +347,7 @@ sub _after_end {
 
 sub _upstream {
     my ($bvf, $feat, $dist) = @_;
+    $dist += get_max_shift_length($bvf);
     return $feat->strand == 1 ? 
         _before_start($bvf, $feat, $dist) : 
         _after_end($bvf, $feat, $dist);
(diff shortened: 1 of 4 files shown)


sarahhunt:
Some tests would be good.

modules/Bio/EnsEMBL/Variation/Utils/VariationEffect.pm
```diff
@@ -347,6 +347,7 @@ sub _after_end {
 
 sub _upstream {
     my ($bvf, $feat, $dist) = @_;
+    $dist += get_max_shift_length($bvf);
```
sarahhunt: is this something that can be cached?

aparton: Unfortunately not, if a new shift hash is created then this value could change, the defensive thing to do is to calculate it every time

dglemos:

PR #531, opened 2020-01-28
stop calculating mfv on the fly

at7:
Removing the check if the species is human ensures that motif_feature_variants are no longer calculated on the fly. https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-2860


ima23:
merged to `master` and `release/100`

PR #535, opened 2020-01-30
Fix script to post process db tables

dglemos:
Fix hash that contains all the citation attributes.


at7:
merged into master and release/100

PR #543, opened 2020-02-20
Support API for SV allele frequency

dglemos:


at7:
Merged into master and release/100

PR #548, opened 2020-02-26
update phenotype dump script for canis_lupus_familiaris - e100

ima23:


scripts/export/dump_phenotypes.pl
```diff
@@ -96,12 +96,16 @@ =head1 CONTACT
...
-  next if ($dbname !~ /^[a-z]+_[a-z]+_variation_\d+_\d+$/i);
+  next if ($dbname !~ /[a-z]+_[a-z]+_variation_\d+_\d+$/i);
   next if ($dbname =~ /^master_schema/ || $dbname =~ /private/);
 
   $dbname =~ /^(.+)_variation_.+_(.+)/;
   my $s_name = $1;
   my $assembly = $2;
+
+  next if ($dbname !~ /^[a-z]+_[a-z]+_variation_\d+_\d+$/i &&
+           $dbname !~ /canis_lupus_familiaris_variation_\d+_\d+$/ );
```
sarahhunt: Lines 106 and 99 are very similar - L99 looks redundant. Can you remove L99 and move the additional dbname checking at L106 up to be with the dbname checking at L100 for clarity?

ima23: @sarahhunt : I agree the lines look similar. L99 is selecting only species looking variation dbs while L106 makes sure that only the actual needed species dbs + dog are further processed and test databases of the form `<user_name>_<binominal_species_name>_variation_`... are skipped. 
Removing line L99 will fail L102 as it will run over all dbs on the server.

ima23: I'll test another version of it.

PR #549, opened 2020-03-02
Task100/human gvf vcf dumps

at7:
Updates which were required for dumping human GVF and VCF files for GRCh37 and GRCh38.

### DIFF
--- a/modules/Bio/EnsEMBL/Variation/Pipeline/ReleaseDataDumps/InitJoinDump.pm
+++ b/modules/Bio/EnsEMBL/Variation/Pipeline/ReleaseDataDumps/InitJoinDump.pm
@@ -62,7 +62,7 @@ sub run {
         # homo_sapiens_structural_variations-chr10.vcf.gz 
         if ($file =~ m/\.($file_type\.gz|$file_type)$/) {
           my $file_name = $file;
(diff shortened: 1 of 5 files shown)


modules/Bio/EnsEMBL/Variation/Pipeline/ReleaseDataDumps/InitJoinDump.pm
```diff
@@ -58,19 +58,19 @@ sub run {
       closedir($dh);
       foreach my $file (@dir_content) {
         next if ($file =~ m/^\./);
-        if ($file =~ m/\.$file_type\.gz/) {
+        # homo_sapiens_structural_variations-chr10.gvf
+        # homo_sapiens_structural_variations-chr10.vcf.gz 
+        if ($file =~ m/\.$file_type\.gz$|\.$file_type/) {
```
aparton: My regular expressions are a little rusty, but wont anything that matches `/\.$file_type/` also match `/\.$file_type.gz/`? I suggest` /\.($file_type\.gz|$file_type)$/` (although I haven't tested it)

aparton:
Thanks. I know I'm being a pain, but we should probably make a similar change to Ln 65

aparton:

PR #555, opened 2020-03-09
Fix somatic flag for SV cosmic import

dglemos:
Some cosmic variants had somatic = 0 after cosmic import. 
Ticket: ENSVAR-2980


at7:
Merged into master and release/100

PR #539, opened 2020-02-12
Support citation sources in API

dglemos:
Changes to API to report citation sources: variation_citation.data_source_attrib

PR #560, opened 2020-03-11
Support citation sources in API (2) 

dglemos:
Duplication of #539


modules/Bio/EnsEMBL/Variation/DBSQL/PublicationAdaptor.pm
```diff
@@ -206,22 +206,72 @@ sub fetch_all_by_dbID_list {
...
+    $sth->bind_columns(\$publication_id, \$data_source_attrib);
     while ($sth->fetch()){
-        push @pub, $self->fetch_by_dbID($publication_id)
+
+      my $publication = $self->fetch_by_dbID($publication_id);
+
+      if($data_source_attrib) {
+
+        if($data_source_attrib =~ /,/) {
+          my @sources = split /,/, $data_source_attrib;
```
at7: a split also works if the string you are planning to split doesn't contain the split character. You don't need to have a separate case for when $data_source_attrib doesn't contain a ','

modules/Bio/EnsEMBL/Variation/DBSQL/PublicationAdaptor.pm
```diff
@@ -206,22 +206,51 @@ sub fetch_all_by_dbID_list {
...
+
+      my $publication = $self->fetch_by_dbID($publication_id);
+
+      if($data_source_attrib) {
+        my @sources = split /,/, $data_source_attrib;
+        foreach my $source_attrib (@sources) {
+          my $source = $attrib_id_to_value->{$source_attrib};
+
+          if(!defined($source)) {
+            my $sth_attrib = $self->prepare(qq{SELECT value FROM attrib WHERE attrib_id = ? });
```
at7: You can use the AttributeAdaptor for getting the value with: attrib_value_for_id(). We use it in the the PhenotypeAdaptor: $self->db->get_AttributeAdaptor->attrib_value_for_id($id). I would suggest to first initialise the adaptor my $attribute_adaptor = $self->db->get_AttributeAdaptor; And then you can use it whenever you need to look up a value: $attribute_adaptor->attrib_value_for_id($data_source_attrib).

modules/Bio/EnsEMBL/Variation/Publication.pm
```diff
@@ -282,4 +292,66 @@ sub variations{
...
+
+  if($variation_id !~ /^\d+$/ || !defined($variation_id)) {
+    throw('variation_id is not valid');
+  }
+
+  if(!defined($source) || !defined($SOURCES{$source})) {
+    throw('Source is not valid');
+  }
+
+  return $self->{'variation_id_to_source'}->{$variation_id}->{$source} = 1;
```
at7: You can remove the return here. It is only a setter method and doesn't return anything.

PR #561, opened 2020-03-11
Set empty authors to NULL in publication table

dglemos:
ENSVAR-2960


dglemos:
This is not the right way to fix this issue. The change will be in another PR.

PR #568, opened 2020-03-19
Citations import fix NULL authors

dglemos:


modules/t/publicationAdaptor.t
```diff
@@ -77,6 +77,21 @@ $va->db->include_failed_variations(1);
 my $varfs = $va->fetch_all_by_publication($pubs->[0]);
 ok(scalar(@{$varfs}) ==2,   "variation count by publication - inc fails");
 
+## store NULL authors
```
helensch: Suggest comment modified to say what testing e.g. authors stored as NULL when not  provided. 
Includes "" and 0.

modules/t/publicationAdaptor.t
```diff
@@ -77,6 +77,21 @@ $va->db->include_failed_variations(1);
...
+my $pub_store_null = Bio::EnsEMBL::Variation::Publication->new(
```
helensch: The test is similar to the other store test. If the database is left intact between test runs, the records are kept.
For future suggest considering using hide and restore e.g. in motifFeatureVariationAdaptor.t

dglemos: You mean using hide and restored for just this new test?

helensch: Yes - it was something to consider for future. - Do we need to do this generally for tests and are DBAdaptors different. I'll add a ticket for.

helensch:

PR #570, opened 2020-03-26
Switch between http and ftp when checking for presence of remote files

azangru:
We are likely to switch Ensembl to retrieving files from ftp server using http protocol rather than ftp. Here is the code to support this change — it checks the protocol of the link to the file, and uses an http client rather than an ftp client if the link uses http protocol.

Related PR in public plugins: https://github.com/Ensembl/public-plugins/pull/320

Related Jira ticket: https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-5607

### DIFF
--- a/modules/Bio/EnsEMBL/Variation/DBSQL/BaseAnnotationAdaptor.pm
+++ b/modules/Bio/EnsEMBL/Variation/DBSQL/BaseAnnotationAdaptor.pm
@@ -59,6 +59,7 @@ package Bio::EnsEMBL::Variation::DBSQL::BaseAnnotationAdaptor;
 use JSON;
 use Cwd;
 use Net::FTP;
+use HTTP::Tiny;
 use URI;
 
 use Bio::EnsEMBL::Utils::Exception qw(throw warning);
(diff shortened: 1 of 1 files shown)


modules/Bio/EnsEMBL/Variation/DBSQL/BaseAnnotationAdaptor.pm
```diff
@@ -279,5 +293,12 @@ sub _ftp_file_exists {
   }
 }
 
+sub _check_file_over_http {
+  my ($self, $uri) = @_;
+  my $ua = HTTP::Tiny->new();
+  my $response = $ua->head($uri);
+
+  return !!$response->{success};
```
at7: one ! too many

azangru: I mostly blindly copied this from https://metacpan.org/release/URL-Exists/source/lib/URL/Exists.pm 😄 
But now that you mention it, it looks like there is no need for the `!!` at all. The subroutine should return true (well, `1`) if the request is successful and false (`0`) if it isn't. I believe that's what the `$response->{success}` field contains anyway ([link to docs](https://metacpan.org/pod/HTTP::Tiny))

at7:
Thank you Andrey. All existing tests pass and I also tested the code in my sandbox and it all works fine. I will add the same functionality to release 100/101 and will also add tests then to bring the test coverage up again.

modules/Bio/EnsEMBL/Variation/DBSQL/BaseAnnotationAdaptor.pm
azangru: I removed the redundant double bang. `$response->{success}` is either 1 or 0.

azangru:
We decided to add a new variable to the DEFAULTS.ini in the public plugins, so that we still use ftp urls in links on our pages, but use the http url to actually access files stored on the ftp server. [PR in public plugins](https://github.com/Ensembl/public-plugins/pull/320) has been updated, and [another PR](https://github.com/Ensembl/ensembl-webcode/pull/763) created in ensembl-webcode.

PR #573, opened 2020-03-28
Add special characters to be replaced

dglemos:
- Special characters are found in publication titles and authors. This PR adds those to the list of characters to be replaced.

- Change citations import to also skip publications with titles like 'P1-200' and to delete title = 'abstract'


scripts/import/import_EPMC.pl
```diff
@@ -674,7 +674,7 @@ sub clean_publications{
...
+    my $wrong_title_sth = $dba->dbc->prepare(qq[ select publication_id, title from publication where title like '%Errata%' or title like '%Erratum%' or title like '%In This Issue%' or title like '%Oral abstracts%' or title like '%Oral Presentations%' or title like '%Proffered paper%' or title like '%Subject Index%' or title like '%Summaries of Key Journal Articles%' or title like '%This Month in The Journal%' or title like 'Index%' or title like '%Table of Contents%' or title like '%Not Available%' or title like 'Beyond Our Pages%' or title like 'EP News%' or title like 'ACTS Abstracts%' or title like 'Poster %' or title like 'Abstracts.%' or title like 'Abstract.%' ]);
```
ima23, line 677: This line would be easier to read if broken down on multiple lines.

dglemos: Thanks for the suggestion. It's done.

ima23:
merged to `master` and `release/101`

PR #574, opened 2020-03-31
Patch publication table 

dglemos:
varchar(255) is not enough for one new publication.authors


helensch:
Will a PR be needed for the test database as well?

dglemos:
Better do that too. I'll open another PR for the test db.

dglemos:
Patch is not going to be applied to postreleasefix/101. It will be against master (e102).

PR #575, opened 2020-03-31
Patch publication table in test database

dglemos:


dglemos:
This is not needed anymore. The patch is going to be applied to master (e102).

PR #578, opened 2020-04-02
Import SV - change sql to insert or update study 

dglemos:
Study table has some values 'NULL' instead of NULL
ENSVAR-3134

- Changed the SQL queries to insert or update a study
- Found HTML character in study description `&lt;` add to `replace_text`

### DIFF
--- a/scripts/import/import_all_sv_data.pl
+++ b/scripts/import/import_all_sv_data.pl
@@ -379,6 +379,10 @@ sub study_table{
   }
   my $external_link = ($display_name{$author}) ? $display_name{$author} : $pubmed;
 
+  if ($external_link && $external_link eq 'NULL' && $external_link eq '') {
+    $external_link = undef;
+  }
+
   # URL
   $study =~ /(\w+\d+)\.?\d*/;
   my $study_ftp = $1;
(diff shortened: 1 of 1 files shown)


scripts/import/import_all_sv_data.pl
```diff
@@ -428,62 +427,45 @@ sub study_table{
...
-        $external_link_sql = '';
-      }
-      else {
-        $external_link_sql = "external_reference='$external_link',";
-      }
-    }
-    else {
-      $external_link_sql = "external_reference='$external_link',";
+    if ($external_link && $external_link eq 'NULL' && $external_link eq '') {
+      $external_link = undef;
```
sarahhunt, line 431: should this be null or empty?

scripts/import/import_all_sv_data.pl
```diff
@@ -428,62 +427,45 @@ sub study_table{
...
+                  );
 
-    $dbVar->do($stmt);
   }
   # INSERT
   else {
-    if ($external_link && $external_link ne 'NULL' && $external_link ne '') {
-      $external_link = "'$external_link'";
+    if($external_link && $external_link eq 'NULL') {
+      $external_link = undef;
```
sarahhunt, line 446: this is similar to L430 - can nulls be dealt with once per input line?

dglemos: Good point. I'll do this before the if .. else

scripts/import/import_all_sv_data.pl
dglemos: It didn't make sense, it's fixed now.

PR #579, opened 2020-04-03
Delete option not used from citations import

dglemos:


sarahhunt:
Line 79 'our $pheno_adaptor...' is no longer needed

PR #580, opened 2020-04-17
update projection script

ima23:
- add reconnect_when_lost
- move init columns into own method

PR for jira ENSVAR-3172.


at7:

PR #581, opened 2020-04-20
Task101/remapping updates

at7:
--add check if remapped variation_feature contains duplicates (https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-2909)
--check that patches are the same between old and new variation database (https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-3126)
--summarise variation_feature counts before and after remapping in separate report file to make any remapping problems more obvious


dglemos:

PR #585, opened 2020-04-24
initial scripts to format Nextstrain JSON to VCF

sarahhunt:


aparton:
I can get json2geno.pl working as I'd expect, looks good.

I'm having a little trouble with geno2vcf.pl though, although it does still generate a result. When I run this, I get a collection of warnings that look like 
`Use of uninitialized value in string ne at nextstrain_geno2vcf.pl line 167.
Ref swap:	3738	C	T
Use of uninitialized value in string ne at nextstrain_geno2vcf.pl line 167.
Ref swap:	3738	C	T
Use of uninitialized value in string ne at nextstrain_geno2vcf.pl line 167.
Ref swap:	3738	C	T
`

This seems to be down to the reference value found in the variant file not being included in the $alt2code hash. 

Additionally, the input order as described in `die "\nUsage: nextstrain_geno2vcf.pl  [genome.fa]  [variants file] \n\n" unless scalar(@ARGV)==2;` is the wrong way round, the script is expecting the variants file first.

sarahhunt:
Thanks @aparton - the ref check reporting and usage line are fixed now. So json2geno makes sense?

PR #586, opened 2020-04-24
Task101/ancestral allele pipeline updates

at7:
I added options to
--only update none dbSNP variants which is useful when we didn't have a new dbSNP import but only updated COSMIC etc
--not to compare stats between releases which is not helpful if ancestral alleles for dbSNP variants haven't changed


dglemos:

PR #587, opened 2020-04-27
Dont check transcript attributes if ENST transcript

aparton:
Skips the attr tests if it's an ENST transcript. 

No test has been added as no output should have changed, should only provide a minor speedup


sarahhunt:
^ENS would be better as it would stop other species going through the checks too.

sarahhunt:
@aparton - we'll need similar for postreleasefix/101, won't we.

PR #590, opened 2020-04-28
update post process script

ima23:
add done print statement to publications section


dglemos:

PR #594, opened 2020-05-19
add configuration for GGVP data

at7:
Configure Gambian Genome Variation Project data. Here is a script which retrieves allele frequencies for a variant: /nfs/production/panda/ensembl/variation/anj/frequencies_from_vcf.pl
It will return frequencies for new GGVP populations.
Config updates have also been tested on my sandbox.


ima23:
merged to `release/101` and `master`

PR #567, opened 2020-03-19
update attribs from production

ima23:

PR #598, opened 2020-05-26
Add attribute

dglemos:
ENSVAR-2550
Following PR: #567


helensch:

PR #601, opened 2020-05-28
spdi - test NT sequence

dglemos:


at7:

PR #603, opened 2020-06-03
Schema update for e102

ima23:
- patch a: schema update 102
- patch b: new clinsig attribute


dglemos:
You should add the patch IDs to the `sql/table.sql` 
https://github.com/Ensembl/ensembl-variation/blob/b32572cf8ba6a6ab3b1ad023f1b453243ee345a0/sql/table.sql#L1835

PR #605, opened 2020-06-03
fixes for clinvar import -e102

ima23:
- fix import script for bug  ENSVAR-3206 (fixed via sql in e101)
- add warning for sanity check of rsID  , ENSVAR-3177


scripts/import/clinvar_post_process.pl
```diff
@@ -119,6 +120,11 @@ sub run_variation_checks {
                     "id"          =>  $l->[0],
             );
 
+        $clinvar_vars{$var{name}} =1;
+        #check for rsID sanity
+        my $number = $var{name}  =~ s/rs//r;
+        warn "WARNING: clinvar rsID less than 650mil, likely typo! $var{name} \n" if $number < 650000000;
```
dglemos, line 126: You could add a comment on why is this number checked `650000000`.

PR #612, opened 2020-06-23
Move decode_text to Utils

dglemos:
Move method decode_text from SV import to Utils so it can be used by other imports and pipelines.


dglemos:
If you're happy with the changes, you can merge it.

PR #614, opened 2020-07-02
Fix PhenotypeAdaptor store for new Phenotypes

ima23:
By default a new Phenotype object is created with the class_attrib (trait) but not the corresponding class_attrib_id. 

The class_attrib_id should be fetch before store if it's not present.


at7:
merged into master and release/102

PR #615, opened 2020-07-03
PhenotypeAdaptor store for new Phenotypes - e101

ima23:
same as https://github.com/Ensembl/ensembl-variation/pull/614

### DIFF
--- a/modules/Bio/EnsEMBL/Variation/DBSQL/PhenotypeAdaptor.pm
+++ b/modules/Bio/EnsEMBL/Variation/DBSQL/PhenotypeAdaptor.pm
@@ -265,6 +265,12 @@ sub get_all_phenotype_class_types {
 sub store{
    my ($self, $pheno) = @_;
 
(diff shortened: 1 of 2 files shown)


modules/Bio/EnsEMBL/Variation/DBSQL/PhenotypeAdaptor.pm
```diff
@@ -265,6 +265,12 @@ sub get_all_phenotype_class_types {
 sub store{
    my ($self, $pheno) = @_;
 
+   # check if phenotype class_attrib_id is present if class_attrib is
```
at7, line 268: Is the comment correct? I think the code checks if class_attrib is present if class_attrib_id is not?
Do you need the comment?

ima23: Thanks for checking. The comment is not clear indeed, have update it.

PR #616, opened 2020-07-06
Prevent start_lost check when shifting out of translateable region

aparton:


sarahhunt:
I looks like there is something odd happening in _overlaps_start_codon and the variant should exit this routine earlier.
cdna_start_unshifted and cdna_end_unshifted are null, as the variant doesn't overlap the cDNA. The shifting offset is added to this, to give a cdna_start & cdna_end of 500, which overlaps the start codon. This addition would only be correct if we knew the the variant was at the base before the cDNA, which we don't.

jamie-m-a:
Looks like this was addressed under https://github.com/Ensembl/ensembl-variation/pull/1167 

Closing this off now.

PR #617, opened 2020-07-06
TV Updates for shifting flag

aparton:


modules/Bio/EnsEMBL/Variation/Pipeline/VariationConsequence_conf.pm
```diff
@@ -95,6 +95,10 @@ sub default_options {
         
         disambiguate_single_nucleotide_alleles => 0,
 
+        # shifting variants within repeated regions in the 3' direction is switched off by default.
+	
+        prevent_shifting => 0,
```
helensch, line 100: From: 
Bio::EnsEMBL::Variation::TranscriptVariation::new 
If --no_shift equals 1, or is undefined, then we don't shift.

If prevent_shifting=0 by default, then -no-shift = 0. Won't this mean that it will shift by default?
Is it intended for shifting to be the default in the pipeline?





If prevent_shifting = 0 by default, then -no_shift = 0.

aparton: Yep, you're absolutely correct, that's just me being a numbskull and setting the default to the wrong value.

helensch: merged into master and release/102

PR #621, opened 2020-07-17
Feature100/support chr synonyms in offline

at7:
This is mainly to support the G2P plugin in offline mode and trying to get additional allele frequencies from the gnomADg v3 VCF files. In the latest gnomADg v3 VCF files chromosomes start with 'chr'.


at7:
I'm trying to get frequencies from a VCFCollection from within the plugin: https://github.com/Ensembl/VEP_plugins/blob/release/100/G2P.pm#L691 At this point the seq_region doesn't include 'chr' and because I'm running offline the VCFCollection cannot get synonyms from the database. I could pass on the synonym file from the cache file to the VCFCollection and then use those synonyms. Which is a bit more work for a very specific use case looking up gnomad v3 here. How would you access the synonym_file from with a plugin?

PR #622, opened 2020-07-20
Task102/remapping

at7:
Only check that the same patches have been applied. Remove table comparison which was causing problems when MTMP tables were present in one but not the other database.

Rename ehive dbname.

### DIFF
--- a/modules/Bio/EnsEMBL/Variation/Pipeline/Remapping/Remapping_conf.pm
+++ b/modules/Bio/EnsEMBL/Variation/Pipeline/Remapping/Remapping_conf.pm
@@ -58,6 +58,7 @@ sub default_options {
         map_to_chrom_only       => 1,
         entries_per_file        => 200000,
         dump_multi_map          => 1,
+        skip_patch_comparison   => 0,
 
         bwa                     => 'bwa',
         samtools                => 'samtools',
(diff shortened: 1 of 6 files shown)


modules/Bio/EnsEMBL/Variation/Pipeline/Remapping/Remapping_conf.pm
```diff
@@ -83,7 +84,7 @@ sub default_options {
             -port   => $self->o('hive_db_port'),
             -user   => $self->o('hive_db_user'),
             -pass   => $self->o('hive_db_password'),            
-            -dbname => $ENV{'USER'} . '_' . $self->o('pipeline_name'),
+            -dbname => $self->o('hive_db_name'),
```
dglemos, line 87: Shouldn't it be the same as the other pipelines? 
`-dbname => $ENV{'USER'} . '_ehive_remapping_' . $self->o('ensembl_release') . '_' . $self->o('assembly') . '_' . $self->o('species'),`

at7:
The dbname is overwritten by the more specific modules: RemappingQTL_conf.pm, RemappingStructuralVariationFeature_conf.pm and RemappingVariationFeature_conf.pm

dglemos:
Merged into release/102 and master

PR #625, opened 2020-07-24
update hive dbname

at7:
I decided not add species and assembly because for all non human species for which we can compute ancestral alleles we can run the pipeline together. We can add more information into the pipeline_name parameter if needed.


aparton:
merged into release/102 and master

PR #626, opened 2020-07-26
update hive dbname

at7:


aparton:
merged into release/102 and master

PR #628, opened 2020-07-29
hgvs_transcript and hgvs_protein now clean up post-shifting

aparton:
Related to https://github.com/Ensembl/ensembl-vep/issues/812

hgvs_transcript and hgvs_protein were shifting HGVS by default, yet not correctly tidying up the new shifted values for subsequence plugin use


modules/t/variation_effect.t
```diff
@@ -1526,6 +1526,27 @@ my $tv_cosmic = Bio::EnsEMBL::Variation::TranscriptVariation->new(
...
+my $tv_clean = Bio::EnsEMBL::Variation::TranscriptVariation->new(
+    -variation_feature  => $vf_clean,
+    -transcript         => $transcript,
+);
+my $tva = $tv->get_all_alternate_BaseVariationFeatureOverlapAlleles();
+
+my $pre_sequence = $tva->[0]->variation_feature_seq;
+$tv->hgvs_transcript();
+my $post_sequence = $tva->[0]->variation_feature_seq;
+is($pre_sequence, $post_sequence, 'hgvs_transcript cleans up variation_feature_seq');
```
at7, line 1548: pre_sequence and post_sequence are the same even without your code update in TranscriptVariationAllele. I would like to see a test case where post_sequence would be different without your code change.

at7:
The cleanup doesn't work and the shift_hash is still being used when the plugins are run. I cannot follow the logic of shift_hash and only have an example where using the hgvs option but without any additional shift parameters is causing different outputs for the Downstream plugin. See my files in PANDA/anja/vep_plugin_issue_342:
--run_vep.sh and results test_small_out.vep
--run_vep_with_hgvs.sh and results test_small_out_with_hgvs.vep
It seems like that the TranscriptVariationAllele::codon sub is still having access to shift_hash.

at7:
Hi @aparton 
I tested this branch with the VEP plugins issue and it now returns the expected results in the Downstream plugin when using the --hgvs option. But could you please check this PR for the reason of the failed test?

diegomscoelho:
Hi @likhitha-surapaneni, I've only notice this PR after opening my fix. It looks like it's simpler to avoid differences in `variation_feature_seq`. Alterations in my PR https://github.com/Ensembl/ensembl-variation/pull/814.

diegomscoelho:
Hi @likhitha-surapaneni, I've only notice this PR after opening my fix. It looks like it's simpler to avoid differences in `variation_feature_seq`. Alterations in my PR https://github.com/Ensembl/ensembl-variation/pull/814.

PR #631, opened 2020-08-10
update gerp file location for 102

at7:
gerp files haven't changed for 102. I only updated the name of the release folder. I also added the gerp file for sheep rambouillet.


ima23:
merged to `master` and `release/102`

PR #632, opened 2020-08-13
add new sheep to dump script - e102

ima23:


at7:
merged into master and release/102

PR #645, opened 2020-09-24
SV import updates

dglemos:
Add source data_types (ENSVAR-3412)
Fix phenotypes with a comma in the description to avoid being split into two different phenotypes.

### DIFF
--- a/scripts/import/import_all_sv_data.pl
+++ b/scripts/import/import_all_sv_data.pl
@@ -1533,19 +1533,18 @@ sub parse_9th_col {
     
     # Phenotype
     if ($key eq 'phenotype') {
-      # The input is a string of multiple phenotypes separated by comma
-      # However, some phenotypes contain a comma in the description
+      # The input is a string of multiple phenotypes separated by comma and space ', '
+      # Some phenotypes contain a comma in the description (without a space)
       # Replace the comma before splitting the multiple phenotypes
       # Example: '46,XX Testicular Disorders of Sex Development'
-      if($value =~ /46,/) {
-        $value =~ s/46,/46-/g;
-      }
 
       $value =~ s/, /__/g;
+      $value =~ s/,/---/g;
+
       foreach my $phe (split(',', $value)) {
         #Change back to comma
-        if($phe =~ /46-/) {
-          $phe =~ s/46-/46,/g;
+        if($phe =~ /---/) {
+          $phe =~ s/---/,/g;
         }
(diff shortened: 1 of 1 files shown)


scripts/import/import_all_sv_data.pl
```diff
@@ -128,18 +128,17 @@ =head1 CONTACT
 
 my %display_name = ( 'COSMIC' => 'https://cancer.sanger.ac.uk/cosmic/' );
 
-my $add             = '';
```
ima23, line 131: The reason for the `$add` was to have an easy way to do a test import on test tables without renaming manually tables in the db. It would be nice to keep this feature in. Ideally this option would be accessible from outside the script as a real option.

scripts/import/import_all_sv_data.pl
```diff
@@ -1530,8 +1532,19 @@ sub parse_9th_col {
     
     # Phenotype
     if ($key eq 'phenotype') {
+      # Some phenotypes contain a comma in the description
+      # Replace the comma before splitting by comma
```
ima23, line 1536: The comment `Replace the comma before splitting by comma` could be clearer. Can you add a note that the input is a string of multiple phenotypes separated by comma?

scripts/import/import_all_sv_data.pl
```diff
@@ -1530,8 +1532,19 @@ sub parse_9th_col {
...
+      # Example: '46,XX Testicular Disorders of Sex Development'
+      if($value =~ /46,/) {
```
ima23, line 1538: Is there a way to catch other cases, or is `46` the only expected? It might not be trivial, maybe a regex to check for a phenotype made out of numbers?

ima23 (approved):
Functionality is there and change fixes the observed issue. Just a few comments for future proofing.

scripts/import/import_all_sv_data.pl
dglemos: I'll change it to make it more clear.

scripts/import/import_all_sv_data.pl
dglemos: Until now `46,` is the only present in the file. I tried to replace all commas in the description but some phenotypes are separated by `,` and others by `, ` making the splitting more difficult.

ima23 (approved):

ima23:
merged to `release/103` and `master`

PR #650, opened 2020-10-13
Add API support for phenotype.stable_id - for e103

ima23:
ParaSite use this column in the phenotype table and getters are required.

### DIFF
--- a/modules/Bio/EnsEMBL/Variation/Phenotype.pm
+++ b/modules/Bio/EnsEMBL/Variation/Phenotype.pm
@@ -82,7 +82,7 @@ sub new {
   my $caller = shift;
   my $class  = ref($caller) || $caller;
   my $self = $class->SUPER::new(@_);
-  my ($dbID, $stable_id, $description, $name, $class_attrib) = rearrange([qw(dbID stable_id DESCRIPTION NAME CLASS_ATTRIB)], @_);
+  my ($dbID, $stable_id, $description, $name, $class_attrib) = rearrange([qw(dbID STABLE_ID DESCRIPTION NAME CLASS_ATTRIB)], @_);
 
   $class_attrib ||= ATTRIB_TRAIT; #default phenotype class type
 
--- a/modules/Bio/EnsEMBL/Variation/Phenotype.pm
+++ b/modules/Bio/EnsEMBL/Variation/Phenotype.pm
@@ -121,7 +121,7 @@ sub dbID {
 
 =head2 stable_id
 
-  Example    : $name = $obj->stable_id()
+  Example    : $stable_id = $obj->stable_id()
   Description: Getter/Setter for the stable_id
   Returntype : string
   Exceptions : none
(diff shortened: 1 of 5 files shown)


ima23:
The updated tests pass: phenotype.t and phenotypeAdaptor.t. 
The other phenotype related tests pass too.
Failed tests due to travis setup on the postreleasefix/103 branch using master branches.

modules/Bio/EnsEMBL/Variation/Phenotype.pm
```diff
@@ -82,12 +82,13 @@ sub new {
   my $caller = shift;
   my $class  = ref($caller) || $caller;
   my $self = $class->SUPER::new(@_);
-  my ($dbID, $description, $name, $class_attrib) = rearrange([qw(dbID DESCRIPTION NAME CLASS_ATTRIB)], @_);
+  my ($dbID, $stable_id, $description, $name, $class_attrib) = rearrange([qw(dbID stable_id DESCRIPTION NAME CLASS_ATTRIB)], @_);
```
aparton, line 85: I know this is mega-neurotic, but it bothers me more than it should that stable_id isn't in caps here

aparton (approved):
Looks good.

modules/Bio/EnsEMBL/Variation/Phenotype.pm
```diff
@@ -118,6 +119,23 @@ sub dbID {
   return $self->{'dbID'};
 }
 
+=head2 stable_id
+
+  Example    : $name = $obj->stable_id()
```
helensch, line 124: Suggest using $stable_id = $obj->stable_id() to avoid confusion with method name example
 $name = $obj->name()

modules/t/phenotypeAdaptor.t
```diff
@@ -57,12 +59,13 @@ ok($p && $p->name eq 'ACH', "fetch_by_description");
 $p->name('test');
 $p->description('test');
 delete $p->{dbID};
-
+delete $p->{stable_id};
```
helensch, line 62: Suggest:
- a test be added to test storing a stable_id
- the store test use hide/restore so that test can be running when keeping test database intact

helensch: The test phenotypeAdaptor.t cannot be re-run when keeping the test database intact due to Duplicate Entries.
Suggest using hide/restore to resolve this - either for this PR or as a separate ticket.

ima23:
Thanks @helensch for checking, the PR is ready for review again.
I have moved the added store test at the end with hide/store.
Had a look and tested how to make the `phenotypeAdaptor.t` run with keeping test db intact but it seems not straight forward as it's legacy test code that requires refactoring. Have added this test to the jira ticket ENSVAR-3116.

helensch (approved):

helensch:
Merged into master and release/103

PR #659, opened 2020-11-13
Update ClinVar import - master / e104

ima23:
- add lookup by canonical spdi
- check all alleles for snps
- print change to warn

### DIFF
--- a/scripts/import/import_clinvar_xml
+++ b/scripts/import/import_clinvar_xml
@@ -1072,7 +1072,7 @@ sub get_variant{
     }
 
     my $location_identifier = $var_feat->location_identifier();
-    warn "search SPDI_location_Str: $location_identifier\n";
+    warn "search SPDI_location_identifier: $location_identifier\n";
 
     my $vfs_from_db = $var_feat_adaptor->fetch_all_by_location_identifier($location_identifier);
     # check for > 1 here
--- a/scripts/import/import_clinvar_xml
+++ b/scripts/import/import_clinvar_xml
@@ -178,7 +178,7 @@ while($reader->read) {
       $omim_source->version($source->version);
       save_source_version($omim_source);
     } else {
-      warn "WARNING: no source version found in input xml therefore no update to db source version.\n";
+      print "WARNING: no source version found in input xml therefore no update to db source version.\n";
     }
   } else {
     my $xml = $reader->readOuterXml;
@@ -206,7 +206,7 @@ sub process_clinvar_set {
 
     ## recover from partial loads with file of accessions already loaded
     if ($done{$record{Acc}}){
-      warn "INFO: skipping $record{Acc}\n";
+      print "INFO: skipping $record{Acc}\n";
       return undef;
     }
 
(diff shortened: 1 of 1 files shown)


scripts/import/import_clinvar_xml
```diff
@@ -1046,19 +1050,73 @@ sub get_variant{
...
-  else {
-    warn "No ref or alt allele or location found\n" if $DEBUG == 1;
-    return undef;
+  } else {
+    warn "no ref or alt or chr or start, skipping lookup by location identifier\n";
   }
+
+  #try to match by SPDI: longest allele format
+  #returns undef for ref==alt SPDIs
+  my $result = get_variant_via_SPDI ($record, $alt_allele);
```
sarahhunt, line 1062: Did this work in your testing?

get_variant_via_SPDI returns an object, an array of objects and a scalar, but only the scalar is caught here. I get an error message:
ERROR: Can't call method "add_evidence_value" without a package or object reference at scripts/import/import_clinvar_xml line 753.

and it's calling  add_evidence_value on the alt allele string '-'

sarahhunt:
Generally, it is easier to spot the error messages you don't expect if you send your print the messages you expect.

ima23:
Thanks @sarahhunt for checking. I've updated the array return and added ref checking.
I'm not sure what you mean by the error messages so I've removed the print/warn edit from this PR. Plan to update logs after I understand.

> Generally, it is easier to spot the error messages you don't expect if you print the messages you expect.

scripts/import/import_clinvar_xml
ima23: I haven't seen that error before. What xml did you test? I can have a look.

sarahhunt: I'm using RCV000001471 - I've put the XML in the ClinVar/testdata/ directory on nfs.

sarahhunt:
Data problems may trigger warnings when the import is running. If the logging messages you expect are sent to standard error, it is harder to see these and react to problems.
Ideally, only send messages you need to act on/fix/rerun for to standard error.

scripts/import/import_clinvar_xml
```diff
@@ -1046,19 +1050,72 @@ sub get_variant{
...
-  else {
-    warn "No ref or alt allele or location found\n" if $DEBUG == 1;
-    return undef;
+  } else {
+    warn "no ref or alt or chr or start, skipping lookup by location identifier\n";
   }
+
+  #try to match by SPDI: longest allele format
+  #returns undef for ref==alt SPDIs
+  my ($var, $var_feats) = get_variant_via_SPDI ($record);
```
sarahhunt, line 1062: Can we return the alt allele here too? (For e!105 if not e!104)

ima23: The method does not change the alt allele. The alt stays the same as before from hgvs.

sarahhunt: If the variant is a deletion, HGVS reports only 'del' so we don't have the actual allele. We can obtain this from SPDI (with care)

PR #665, opened 2020-11-26
Update phenotype  import -e103

ima23:
- skip sheep animalqtl as different assembly to ensembl


at7 (approved):

at7:
merged into master and release/103

PR #666, opened 2020-12-01
Fix needed for pig AnimalQTL data

ima23:
ENSVAR-3764


aparton:
Looks good. The inconsistent capitalisation of hash terms could cause us a problem in the future, however that's not within the scope of this PR, so I'm happy to merge.

aparton:
merged into release/103 and master

ima23:
Thanks

> Looks good. The inconsistent capitalisation of hash terms could cause us a problem in the future, however that's not within the scope of this PR, so I'm happy to merge.

Thanks @aparton. On the hash terms capitalisation, these are the ones from the input file, there is an example in the jira ticket.

PR #672, opened 2020-12-10
Revert fix coord system 

dglemos:
Revert PR #667 
This PR fixes failing test in ensembl-vep Parser_HGVS.t`

PR #673, opened 2020-12-10
Revert fix coord system

dglemos:
Same as #672


at7 (approved):
This PR is needed for now in order to pass ensembl-vep tests before branching. An updated version will be submitted which will also increase the coverage

at7:

PR #677, opened 2021-01-08
add dbNSFP version 4.1a

at7:
Add support for dbNSFP file version 4.1a and also throw an error if no column names have been specified for the provided dbNSFP version. We had to start defining the column names for each new dbNSFP file version because there have been changes in the past. It seems more stable now and I need to look into a way of dealing with this better rather than hard coding  the column names for each new file release.


at7:
@ima23 I will add some test data. I let you know when the PR is ready to be reviewed. Thank you

at7:
@ima23 I added the tests and the PR can now be reviewed.

modules/Bio/EnsEMBL/Variation/Utils/DbNSFPProteinFunctionAnnotation.pm
```diff
@@ -142,6 +142,29 @@ my $column_names = {
         pos => 'pos(1-based)'
       },
     },
+  },
+  '4.1a' => {
+    assembly_unspecific => {
```
ima23, line 147: The hash keys could all be written with or all without quotes.

ima23 (approved):
Thanks for adding the clear test data. I agree that it might be worth reviewing if there is even a better way to do this hardcoding, potentially reducing the hardcoding to only contain the columns that changed. 
Happy for this to go in as is and was run for production.

ima23:
merged to `master` and `release/104`

PR #676, opened 2021-01-05
Use AC,AN before AF in VCF freq calc to fix rounding

helensch:
PR for ENSVAR-3763

PR will be made on postrealeasefix/104 after approval/merge of this PR

PR #680, opened 2021-01-13
Use AC,AN before AF in VCF freq calc to fix rounding

helensch:
For e104. See PR for e103 #676
See ENSVAR-3763


aparton (approved):

aparton:
merged into postreleasefix/104 and master

PR #681, opened 2021-01-14
Prf104/ancestral allele pipeline

at7:
-- rename option
-- If the pipeline only updates non-dbSNP variants we only need to compare ancestral allele assignments, before and after the pipeline was run, for non-dbSNP variants

### DIFF
--- a/modules/Bio/EnsEMBL/Variation/Pipeline/AncestralAlleles/Init.pm
+++ b/modules/Bio/EnsEMBL/Variation/Pipeline/AncestralAlleles/Init.pm
@@ -70,9 +70,9 @@ sub run {
   foreach my $species_name (keys %$run_pipeline) {
     my $species_dir = $self->create_species_dir($species_name);
     if ($self->param('create_stats')) {
-      $self->store_previous_release_stats($species_name, $species_dir);
+      $self->store_previous_release_stats($species_name, $species_dir, $self->param('non_dbSNP_only'));
     }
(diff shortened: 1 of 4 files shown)


modules/Bio/EnsEMBL/Variation/Pipeline/AncestralAlleles/Init.pm
```diff
@@ -107,11 +107,17 @@ sub store_previous_release_stats {
   my $self = shift;
   my $species_name = shift;
   my $species_dir = shift;
+  my $non_dbSNP_only = shift;
+  my $sql = qq/SELECT ancestral_allele, COUNT(*) FROM variation_feature GROUP BY ancestral_allele;/;
+  if ($non_dbSNP_only) {
+    $sql = qq/SELECT ancestral_allele, COUNT(*) FROM variation_feature WHERE source_id != 1 GROUP BY ancestral_allele;/;
```
dglemos, line 113: In nomascus_leucogenys_variation_104_3, dbSNP source_id = 2.
Maybe it would be better to fetch the source id.

dglemos:
Coverage goes down because of new function [get_source_id](https://coveralls.io/builds/36828127/source?filename=modules%2FBio%2FEnsEMBL%2FVariation%2FPipeline%2FBaseVariationProcess.pm#L33). I'll merge the PR as it is because there are no tests for /Variation/Pipeline/.

dglemos (approved):

dglemos:
Merged into release/104 and master

PR #683, opened 2021-01-14
Task104/update ehive db name

at7:
Let user be able to overwrite the hive db name.


ima23 (approved):

ima23:
merged to `release/104` and `master`

PR #686, opened 2021-01-21
SPDI uses the sequence version number

dglemos:
Mouse `NC_000068.8:3115317:T:C` fetches the sequence from GRCm38 instead of GRCm39 because fetch_by_spdi_notation doesn't use the sequence version. 

Ticket: ENSVAR-3947


at7:
Could you please at least add a test where the version number is used [here](https://github.com/Ensembl/ensembl-variation/blob/release/102/modules/t/variationFeatureAdaptor.t#L447). A human example works here too. At the moment the [line](https://coveralls.io/builds/36494378/source?filename=modules%2FBio%2FEnsEMBL%2FVariation%2FDBSQL%2FVariationFeatureAdaptor.pm#L2497) is not tested. Maybe you could add a version number to this NT_ [here](https://github.com/Ensembl/ensembl-variation/blob/release/102/modules/t/variationFeatureAdaptor.t#L508)?

dglemos:
Thank you for checking, the test needed an update. The line seems to be tested now.

at7 (approved):

PR #687, opened 2021-01-21
SPDI uses the sequence version number

dglemos:
Same as #686 
Ticket: ENSVAR-3947


at7:
Could you please update the test here too?

at7 (approved):

at7:

PR #694, opened 2021-02-09
SQL Foreign keys - remove variation_citation attrib_id 

dglemos:
The FK datacheck picks up the relationship from sql/foreign_keys.sql. The DCs end up doing a JOIN between a value (attrib.attrib_id) and a set value (variation_citation.data_source_attrib) which is incorrect. 
Related tickets: ENSVAR-1996 and ENSVAR-3980


helensch (approved):

helensch:

PR #699, opened 2021-02-17
VariationFeature - get variant synonyms 

dglemos:
ENSVAR-2692


modules/Bio/EnsEMBL/Variation/Variation.pm
```diff
@@ -433,6 +433,32 @@ sub get_all_synonyms {
...
+  Description: Retrieves a list of all the variation synonyms of this
+               Variation. The output includes the source and the synonyms.
+  Returntype : Listref of strings
+  Exceptions : none
+  Caller     : general
+  Status     : Stable
+
+=cut
+
+sub get_variation_synonyms {
```
aparton, line 449: We already have the subroutines `get_all_synonyms` and `get_all_synonym_sources`, do we need this extra one?

modules/t/variation.t
```diff
@@ -138,6 +138,18 @@ ok($sources[0] eq 'TSC' &&
...
+my $variation_synonyms_1 = $var_syn_1->get_variation_synonyms();
+ok(@$variation_synonyms_1[0] eq 'LSDB: NM_000492.3:c.2125C>T', 'get_variation_synonyms valid (source LSDB)');
+
+my $var_syn_2 = $variation_adaptor->fetch_by_name('rs2299222');
+my $variation_synonyms_2 = $var_syn_2->get_variation_synonyms();
+ok(@$variation_synonyms_2[0] eq 'Archive dbSNP: rs60739517,rs17765152', 'get_variation_synonyms valid (source Archive dbSNP)');
+
+my $var_syn_3 = $variation_adaptor->fetch_by_name('rs117161559');
+my $variation_synonyms_3 = $var_syn_3->get_variation_synonyms();
+ok(@$variation_synonyms_3 == 0, 'get_variation_synonyms - no variant synonyms');
```
aparton, line 152: Is it worth adding a test where there are synonyms from multiple sources?

dglemos:
@aparton Thanks for the comments. I updated the get_all_synonyms() to return the variation synonyms (including the source).

aparton (commented):
Looks good, my only concern is that running in 'include_source' mode requires an empty string to be provided rather than an undefined value. Can we add a `||= '';`?

aparton:

PR #701, opened 2021-03-03
Update SV import - improve phenotypes and add post-processing checks

dglemos:
1. skip non-reported phenotypes (ENSVAR-3995) 
2. add class_attrib_id to phenotypes and report which clinical significance is not supported (ENSVAR-3581)
3. after import check if there are any phenotype entries without phenotype feature

### DIFF
--- a/scripts/import/import_all_sv_data.pl
+++ b/scripts/import/import_all_sv_data.pl
@@ -241,8 +241,8 @@ foreach my $in_file (@files) {
   my $attrib_id;
   if($phenotype_data) {
     $attrib_id = get_attrib_id();
+    phenotype_feature($attrib_id);
   }
-  phenotype_feature($attrib_id);
 
   drop_tmp_table() if (!defined($debug));
   debug(localtime()." Done!\n");
@@ -970,7 +970,7 @@ sub get_attrib_id {
   $stmt->execute() || die;
   my $attrib_id = ($stmt->fetchrow_array)[0];
 
-  debug(localtime()." No attribute 'trait' was found in attrib table!\n") unless defined $attrib_id;
+  die(localtime()." ERROR: No attribute 'trait' was found in attrib table!\nCleanup the tmp tables 'tmp_sv' and 'tmp_sv_phenotype' before importing again.\n") unless defined $attrib_id;
 
   return $attrib_id;
 }
@@ -1584,7 +1584,10 @@ sub parse_9th_col {
 
         $phe =~ s/__/, /g;
 
+        # $phenotype_data is set to 1 when there's valid phenotypes
+        # the attrib_id is going to be fetched only if we have valid phenotype data (#L242)
         $phenotype_data = 1;
+        # if a phenotype is invalid we do not store it and set $skip_phenotype to 1 to skip it in the next lines (#L1594-1595)
         $info->{phenotype}{$phe} = 1 unless $skip_phenotype;
       }
     }
--- a/scripts/import/import_all_sv_data.pl
+++ b/scripts/import/import_all_sv_data.pl
@@ -6,7 +6,7 @@
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
 # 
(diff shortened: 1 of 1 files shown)


scripts/import/import_all_sv_data.pl
```diff
@@ -235,7 +236,14 @@ =head1 CONTACT
...
   structural_variation_set() if ($var_set_id);
   structural_variation_sample();
-  phenotype_feature();
+
+  # Fetch the default attrib id (only if there's phenotype data)
+  my $attrib_id;
+  if($phenotype_data) {
+    $attrib_id = get_attrib_id();
+  }
+  phenotype_feature($attrib_id);
```
ima23, line 245: Why is phenotype_feature($attrib_id) not run only when there is $phenotype_data?

scripts/import/import_all_sv_data.pl
```diff
@@ -955,7 +963,21 @@ sub get_gender {
...
+
+  return $attrib_id;
+}
+
 sub phenotype_feature {
+  my $class_attrib_id = shift;
+
   my $stmt;
   
   # Check if there is some phenotype entries
```
ima23, line 983: Can the input parameter be checked that is defined?  $class_attrib_id

scripts/import/import_all_sv_data.pl
```diff
@@ -955,7 +963,21 @@ sub get_gender {
...
+# Get the phenotype attrib id
+# By default, attrib = 'trait'
+sub get_attrib_id {
+  my $stmt = $dbVar->prepare(qq { SELECT attrib_id FROM attrib WHERE value = 'trait' });
+  $stmt->execute() || die;
+  my $attrib_id = ($stmt->fetchrow_array)[0];
+
+  debug(localtime()." No attribute 'trait' was found in attrib table!\n") unless defined $attrib_id;
+
+  return $attrib_id;
```
ima23, line 975: Should an undef $attrib_id be returned and the script to proceed?

scripts/import/import_all_sv_data.pl
```diff
@@ -1554,11 +1583,13 @@ sub parse_9th_col {
         $phe = replace_char($phe);
 
         $phe =~ s/__/, /g;
-        $info->{phenotype}{$phe} = 1;
+
+        $phenotype_data = 1;
```
ima23, line 1587: Can a comment be added for the meaning of `$phenotype_data =1` and value of `$info->{phenotype}{$phe}`  if `$skip_phenotypes =1`? If the phenotype is to be skipped, why is important to have $phenotype_data =1?

scripts/import/import_all_sv_data.pl
dglemos: It was just to avoid fetching the id when it's not going to be used.

ima23 (commented):
Overall looks good. Mainly a few comments would improve the understanding of the changed. For future it would be easier to review if each resolved ticket would be a commit and any extra small code updates another commit then all in one.

scripts/import/import_all_sv_data.pl
dglemos: I chose to return an undef attrib_id and proceed with the import because the missing attrib_id it's not critical for the import and can be fixed afterwards. 
If I stop the import, then I'll need to do some cleaning (tmp tables) before importing again.

scripts/import/import_all_sv_data.pl
dglemos: get_attrib_id L968 dies if attrib_id is undef.

scripts/import/import_all_sv_data.pl
dglemos: It's better to kill the import and fix the attrib before. I wrote an error that also explains some cleaning has to be done before importing again.

scripts/import/import_all_sv_data.pl
dglemos: Let me know if the comments are not clear.

ima23 (approved):

PR #707, opened 2021-04-14
Bugfix104/fix docs population display 2

at7:
-- for two new EVA species the population_display_group is stored in the vcf config file and overwrites the label
-- also support case where label looks like PRJEB...
-- we used the wrong display group id for nile tilapia, after fix still all works on my [sandbox](http://ves-hx2-76.ebi.ac.uk:5040/Oreochromis_niloticus/Variation/Population?db=core;r=LG19:18792028-18872052;v=LG19:18819865:A_G:PRJEB38548;vdb=variation;vf=LG19:18819865:A_G:PRJEB38548); I create a PR for public-plugins too


aparton:

PR #709, opened 2021-04-19
ClinVar import - fetch more pubmed IDs

dglemos:
ENSVAR-4040

### DIFF
--- a/scripts/import/import_clinvar_xml
+++ b/scripts/import/import_clinvar_xml
@@ -96,6 +96,7 @@ my $slice_adaptor         = $reg->get_adaptor('homo_sapiens', 'core', 'slice');
 $slice_adaptor->dbc->reconnect_when_lost(1);
 
 ## fetch ClinVar source object and update version based on input file name
+# fetch ClinVar and OMIM and updates version for both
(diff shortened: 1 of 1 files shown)


scripts/import/import_clinvar_xml
```diff
@@ -605,6 +608,24 @@ sub get_citations{
       }
     }
   }
+
+  my $observed_in_ca = to_array($structure_ca);
+  foreach my $observed_in (@{$observed_in_ca}) {
+
+    my $observed_data = to_array($observed_in->{ClinicalSignificance});
+    foreach my $obs(@{$observed_data}){
```
sarahhunt, line 616: This provides the results we need, but leaves us with duplicated code. 
It would be neater to call the sub twice with a key to say what field to extract data from, then merge results to avoid the duplication.

Something like 
  $record{citations} = get_citations($set->{ReferenceClinVarAssertion}->{ObservedIn}, "ReferenceClinVarAssertion");
  $extra_citations = get_citations($set->{ClinVarAssertion}, ,"ClinVarAssertion");
push (@{$record{citations}} , @{$extra_citations})

PR #710, opened 2021-04-21
added extra_mem resource class to Remapping_conf.pm

brunocontrerasmoreira:
HI @at7 , while remapping the previous variation to the new contiguous barley (_Hordeum vulgare_) assembly all run_mapping jobs ran out of memory. Selecting the extra_mem resource class solve the problem.


at7:

PR #717, opened 2021-05-10
Report variant synonyms

dglemos:
Same as #699. 
This PR fixes the test failing in perl 5.26


aparton:

PR #718, opened 2021-05-13
Update variation synonyms test

dglemos:
The variation synonyms test fails in Perl 5.26 because the synonyms output order is different  `Archive dbSNP: rs17765152,rs60739517` instead of `Archive dbSNP: rs60739517,rs17765152` (other Perl versions) . To avoid this failure, need to update the test.


aparton:

PR #719, opened 2021-06-01
Task105/remapping pipeline updates

at7:
-- path and queue updates for new codon cluster
-- for remapping pipeline:
   -- remove resource_classes from RemappingVariationFeature_conf.pm, as it is inherited from Remapping_conf.pm
   -- when copying over seq region data from the core to variation database also include coord_system_id

### DIFF
--- a/modules/Bio/EnsEMBL/Variation/Pipeline/AncestralAlleles/AncestralAlleles_conf.pm
+++ b/modules/Bio/EnsEMBL/Variation/Pipeline/AncestralAlleles/AncestralAlleles_conf.pm
@@ -63,8 +63,8 @@ sub default_options {
         hive_no_init => 0,
 
         # the location of your checkout of the ensembl API (the hive looks for SQL files here)
-        hive_root_dir           => '/hps/software/users/ensembl/repositories/' . $ENV{'USER'} . '/ensembl-hive',
-        ensembl_cvs_root_dir    => '/hps/software/users/ensembl/repositories/' . $ENV{'USER'},
+        ensembl_cvs_root_dir    => $ENV{'ENSEMBL_CVS_ROOT_DIR'} || '/hps/software/users/ensembl/repositories/' . $ENV{'USER'},
+        hive_root_dir           => $self->o('ensembl_cvs_root_dir') . '/ensembl-hive',
 
         # a name for your pipeline (will also be used in the name of the hive database)
         pipeline_name           => 'ancestral_alleles',
--- a/modules/Bio/EnsEMBL/Variation/Pipeline/AncestralAlleles/AncestralAlleles_conf.pm
+++ b/modules/Bio/EnsEMBL/Variation/Pipeline/AncestralAlleles/AncestralAlleles_conf.pm
@@ -61,17 +61,16 @@ sub default_options {
         hive_use_triggers => 0,
         hive_auto_rebalance_semaphores => 0, 
         hive_no_init => 0,
+
         # the location of your checkout of the ensembl API (the hive looks for SQL files here)
-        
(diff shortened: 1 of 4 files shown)


helensch (commented):
For setting of hive_root_dir and ensembl_cvs_root_dir, instead of specifying path could something similar to what in EnsemblGeneric_conf.pm be used: i.e.
`'ensembl_cvs_root_dir'  => $ENV{'ENSEMBL_CVS_ROOT_DIR'} || $self->o('ensembl_cvs_root_dir')`

With above suggest using $ENV{'ENSEMBL_ROOT_DIR'} if set in new environment

modules/Bio/EnsEMBL/Variation/Pipeline/AncestralAlleles/AncestralAlleles_conf.pm
```diff
@@ -61,17 +61,16 @@ sub default_options {
         hive_use_triggers => 0,
         hive_auto_rebalance_semaphores => 0, 
         hive_no_init => 0,
+
         # the location of your checkout of the ensembl API (the hive looks for SQL files here)
-        
-        ensembl_cvs_root_dir    => $ENV{'HOME'} . '/bin',
-        hive_root_dir           => $ENV{'HOME'} . '/bin/ensembl-hive', 
+        ensembl_cvs_root_dir    => $ENV{'ENSEMBL_CVS_ROOT_DIR'} || '/hps/software/users/ensembl/repositories/' . $ENV{'USER'},
```
helensch, line 66: Could $self->o('ensembl_cvs_root_dir') be used instead of file path.
Suggest maybe using $ENV{'ENSEMBL_ROOT_DIR'} in future as used in "Codon QuickStart: Running Pipelines"

modules/Bio/EnsEMBL/Variation/Pipeline/Remapping/PreRunChecks.pm
```diff
@@ -60,8 +60,8 @@ sub run {
 
   my $core_dbname = $cdba->dbc->dbname;
   $vdba->dbc()->sql_helper()->execute_update(-SQL => "DELETE FROM seq_region;", -PARAMS => [] );
-  my $sql = "INSERT INTO seq_region(seq_region_id, name)"
-    . "SELECT sr.seq_region_id, sr.name "
+  my $sql = "INSERT INTO seq_region(seq_region_id, name, coord_system_id) "
+    . "SELECT sr.seq_region_id, sr.name, sr.coord_system_id "
```
helensch, line 64: Does the variation.coord_system table need to be updated?

at7: The data checks require that coord_system_id is also populated.

helensch: If new coord_system_id are added to variation.seq_region.coord_system_id, then new records will need to be added to variation_coord_system table. It is the updates to this table that is I was querying.

at7: That is right. But this is done by the scrip ensembl-variation/scripts/misc/update_seq_region_coord.pl. I'm adding the coord_system_id here now because the coord_system_id cannot be NULL and doesn't have a default value. This didn't cause a problem before, but does now on codon.

helensch: Thanks for the explanation.

helensch (approved):

PR #726, opened 2021-06-03
Citations import - format authors

dglemos:
Format authors to store the first 4 authors. If longer than 4, include 'et al'. 
ENSVAR-4182

### DIFF
--- a/scripts/import/import_EPMC.pl
+++ b/scripts/import/import_EPMC.pl
@@ -362,8 +362,8 @@ sub format_authors{
 
   if(defined $authors_final){
     my @author_list = split /, /, $authors_final;
-    if(scalar @author_list > 6) {
-      $new_authors = join(', ', @author_list[0..5]) . ', et al';
+    if(scalar @author_list > 4) {
+      $new_authors = join(', ', @author_list[0..3]) . ', et al';
     }
   }
 
--- a/scripts/import/import_EPMC.pl
+++ b/scripts/import/import_EPMC.pl
@@ -354,7 +354,7 @@ sub get_publication_info_from_epmc{
 }
 
 # Format authors list
-# only store the first 6 authors
+# only store the first 4 authors
 sub format_authors{
   my $authors_final = shift;
 
--- a/scripts/import/import_EPMC.pl
+++ b/scripts/import/import_EPMC.pl
@@ -347,27 +347,27 @@ sub get_publication_info_from_epmc{
     }
 
     ### format authors list
-    my $new_authors = format_authors($ref->{resultList}->{result}->{authorString});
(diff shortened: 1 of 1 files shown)


scripts/import/import_EPMC.pl
```diff
@@ -346,9 +346,30 @@ sub get_publication_info_from_epmc{
...
+    ### format authors list
+    my $new_authors = format_authors($ref->{resultList}->{result}->{authorString});
+    $ref->{resultList}->{result}->{authorString} = $new_authors unless !$new_authors;
+
     return $ref;
 }
 
+# Format authors list
+# only store the first 4 authors
+sub format_authors{
```
at7, line 358: I would be more precise here and call it trim_author_list

scripts/import/import_EPMC.pl
```diff
@@ -346,9 +346,30 @@ sub get_publication_info_from_epmc{
          return undef;
     }
 
+    ### format authors list
+    my $new_authors = format_authors($ref->{resultList}->{result}->{authorString});
```
at7, line 350: instead of new_authors maybe call it trimmed_author_list

scripts/import/import_EPMC.pl
```diff
@@ -346,9 +346,30 @@ sub get_publication_info_from_epmc{
...
+    $ref->{resultList}->{result}->{authorString} = $new_authors unless !$new_authors;
```
at7, line 351: I would prefer: 
$ref->{resultList}->{result}->{authorString} = $trimmed_author_list if $trimmed_author_list;

scripts/import/import_EPMC.pl
```diff
@@ -511,6 +532,10 @@ sub check_dbSNP{
             print $error_log "WARN dbSNP data:\t$l->[1]\t($ref->{resultList}->{result}->{title}) - species not mentioned\n"
         }
 
+        # format authors list
+        my $new_authors = format_authors($ref->{resultList}->{result}->{authorString});
```
at7, line 536: same comments as above on method and variable naming

scripts/import/import_EPMC.pl
```diff
@@ -981,6 +1006,10 @@ sub parse_UCSC_file{
         my ($rs, $pmid, $section, $doi, $title, $authors, $year, $extId ) = split/\t/;
         next if $section =~/refs|ack/; ## not in this publication
 
+        # format authors list
+        my $new_authors = format_authors($authors);
```
at7, line 1010: same comments as above on method and variable naming

dglemos:
@at7 thanks for the suggestions. I've renamed the variables.

at7 (approved):

at7:

PR #731, opened 2021-06-11
Update variant synonyms import with new terms

dglemos:
UniProt file (https://www.uniprot.org/docs/humsavar) has different terms for variant classification. 
ENSVAR-4191


at7 (approved):

at7:

PR #732, opened 2021-06-11
Add special charaters to replace_hex()

dglemos:
These are special charaters found in publications.


at7 (approved):

at7:

PR #733, opened 2021-06-23
Phenotype Annotation Fixes for strict SQL

helensch:
The MySQL servers are using strict SQL.
This fixes SQL errors raised

- extra commas in insert to set values
- truncation of string larger than table.field where stored

For further detail see: ENSVAR-4193

In the shorter term tests should be added to trim_value_string and moved to a utility module
In the longer term the fields being truncated will be remodelled


sarahhunt:
merged to master and release/105

PR #735, opened 2021-07-06
Task105/prepare new db nsfp file

at7:
Combine all steps that are needed to create tabixed dbNSFP files for GRCh37 and GRCh38 in a script which can be run after a new dbNSFP version has been released.


olaaustine (approved):
Tested on the version 4.1a and the script works.

at7:

PR #737, opened 2021-07-06
Adds SpliceRegion Consequences to main VEP

aparton:
Adds the new consequences:
- splice_donor_5th_base_variant (SO:0001787) 
- splice_donor_region_variant (SO:0002170) 
- splice_polypyrimidine_tract_variant (SO:0002169 )
into the main VEP code. Currently, they can be added through the SpliceRegion plugin.

### DIFF
--- a/modules/Bio/EnsEMBL/Variation/BaseTranscriptVariationAllele.pm
+++ b/modules/Bio/EnsEMBL/Variation/BaseTranscriptVariationAllele.pm
@@ -149,11 +149,14 @@ sub _intron_effects {
           $intron_effects->{intronic} = 1;
         }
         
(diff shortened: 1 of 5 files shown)


olaaustine (approved):

sarahhunt:
I am seeing a difference in how this one is annotated:
2	222202192	rs771419226	T	TA

Consequence: splice_region_variant,intron_variant
HGVS: ENST00000350526.9:c.1174-3dup

The last base of the polypyrimidine_tract is duplicated, so the bases -17 to -3 are changed. We previously assigned splice_polypyrimidine_tract_variant. We should talk about this one

PR #738, opened 2021-07-16
after removing empty populations remove check for parus major

at7:
The bug fix can be reverted now that the empty populations have been remove from the parus major database.
https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-4056


aparton:
merged into release/105 and master

PR #741, opened 2021-07-28
Update Structural variants import (strict SQL)

dglemos:
SV import was inserting duplicated values in SET `clinical_significance`. This is not possible anymore because the servers are using strict SQL.

### DIFF
--- a/scripts/import/import_all_sv_data.pl
+++ b/scripts/import/import_all_sv_data.pl
@@ -1854,6 +1854,8 @@ sub post_processing_clinical_significance {
   $sth_stmt->execute();
   my $all_clin_sign = $sth_stmt->fetchall_arrayref();
 
+  my $sth_insert_temp = $dbVar->prepare(qq{ INSERT INTO $temp_clin_table (sv_id, clin_sign) values (?, ?) });
+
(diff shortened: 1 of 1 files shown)


scripts/import/import_all_sv_data.pl
```diff
@@ -1854,7 +1850,24 @@ sub post_processing_clinical_significance {
...
+    my $clin_sign_dup = $data->[1];
+    my @clin_sign_list = split ',', $clin_sign_dup;
+    my %unique_clin_sign;
+    foreach my $key (@clin_sign_list) {
+      $unique_clin_sign{$key} = 1;
+    }
+    my @new_clin_sign_unique = keys %unique_clin_sign;
+    my $clin_sign_unique_final = join ',', @new_clin_sign_unique;
+
+    my $sth_insert_temp = $dbVar->prepare(qq{ INSERT INTO $temp_clin_table (sv_id, clin_sign) values (?, ?) });
```
at7, line 1868: The $sth_insert_temp can be outside of the loop. It doesn't need to newly initialised every time, but can be reused. Just add it before line 1857

at7 (approved):

at7:

PR #743, opened 2021-07-29
Update Mastermind import (private db)

dglemos:
For pribate db grch38: update script to use HGVS grch38. 
More info: ENSVAR-4289


at7 (approved):

at7:

PR #744, opened 2021-07-30
Fix phenotype import for rat and add boxer

dglemos:
Fix rat assembly: ENSVAR-4298
Add canis_lupus_familiarisboxer to pipeline


at7 (approved):

at7:

PR #746, opened 2021-08-20
Update citation import - truncate title

dglemos:
publication.title max size is 300 varchar. 
Update script to truncate title if length >300
ENSVAR-4314

### DIFF
--- a/scripts/import/import_EPMC.pl
+++ b/scripts/import/import_EPMC.pl
@@ -268,10 +268,21 @@ sub import_citations{
         }
         else{
             ## add new publication
+	    my $new_title = $title;
+
+            ## check title size
+            if(length($title) >= 300){
+              # truncate title
+              my $aux = substr($title, 0, 296);
+              my @list_title = split(' ', $aux);
+              pop @list_title;
+              $new_title = join(' ', @list_title);
+              $new_title .= '...';
+            }
 
             ### create new object
             my $publication = Bio::EnsEMBL::Variation::Publication->new(
-                -title    => $title,
+                -title    => $new_title,
                 -authors  => $ref->{resultList}->{result}->{authorString}   || $data->{$pub}->{authors},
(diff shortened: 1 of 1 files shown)


scripts/import/import_EPMC.pl
```diff
@@ -268,10 +268,21 @@ sub import_citations{
         }
         else{
             ## add new publication
+	    my $new_title = $title;
```
helensch, line 271: Could a regexp be used to remove the last characters preceded by a space instead of splitting and joining?

helensch (commented):
Looks fine. Minor suggestions re removing trailing characters and documenting header change.

scripts/import/import_EPMC.pl
dglemos: I tried that, however if title is very long the last occurrence of a space is after the 300 characters. Splittting and joining was the only way I found to add characters until 300 characters and ignore everything else.

scripts/import/import_EPMC.pl
```diff
@@ -386,14 +397,17 @@ sub get_epmc_data{
...
-    my $request  = $server . $id;
+    my $request  = $server . $id . '&format=xml';
 
     my %data;
 
     print "Looking for $request\n\n"  if $DEBUG == 1;
     my $response = $http->get($request, {
-        headers => { 'Content-type' => 'application/xml' }
-                              });
+        headers => {
```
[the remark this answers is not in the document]
dglemos: Thanks for the suggestion. Will add a note on confluence.

scripts/import/import_EPMC.pl
helensch: The suggestion was to use a regexp to remove the last characters preceded by a space instead of splitting and joining after the truncation to 296 characters.  As is the array works, this is fine.  A ticket can be added to review use of regexp for truncation as done in trim_value_string in BasePhenotypeAnnotation.pm

helensch (approved):

PR #750, opened 2021-09-06
change to using DBConnection

at7:
Script didn't work on codon with just `DBI->connect...`. Replacing it with Bio::EnsEMBL::DBSQL::DBConnection is working.


marcoooo (commented):
Problem solved for the `LOAD DATA LOCAL INFILE` failure on CODON

at7:

PR #751, opened 2021-09-07
change to using DBConnection

at7:
Using DBI->connect and load local infile is not working on codon. Replacing it with Bio::EnsEMBL::DBSQL::DBConnection fixes the problem.


at7:

PR #759, opened 2021-09-28
VEP examples generation - check slice exists

helensch:
Fix for release/105
When approved PR will be made for postreleasefix/106


dglemos:
Tested for species _candida_auris_gca_001189475. This species doesn't have examples on the live site e104. 
The code update fix the error message - no examples are generated.

dglemos (approved):

PR #763, opened 2021-10-01
[Deprecated] Change mart_variation_effect.pl to use DBConnection

luca-drf:
Use `Bio::EnsEMBL::DBSQL::DBConnection` instead of `DBI` when loading from a file dump. Similar to #751


luca-drf:
Closed as replaced by #764 and #765

PR #765, opened 2021-10-01
Change mart_variation_effect.pl to use DBConnection

luca-drf:
Use `Bio::EnsEMBL::DBSQL::DBConnection` instead of `DBI` when loading from a file dump. Similar to #751


dglemos (approved):

dglemos:
Merged into release/106 and master

PR #769, opened 2021-11-23
Fetch consequences from VCF INFO

likhitha-surapaneni:
Issue reported: ENSVAR-869


jamie-m-a (commented):
Looks good, tested and working fine. Just wondering if for the changes to VCFcollection.pm you might consider while rather than foreach for some of the nested loops, potentially some of these slices could contain a lot of elements and while is more memory efficient (would need to check impact on speed, though I don't expect there would be much).

PR #770, opened 2021-12-06
Fix ld_vcf to for multiple files and region

helensch:
See ENSVAR-4549


dglemos:
Code tested on codon after setting up HTSLIB_DIR. The two failing tests pass after the code change.

dglemos (approved):

dglemos:
Merged into release/106 and master

PR #772, opened 2021-12-17
Structural variants import - fix tables name in message

dglemos:
Update message to display correct temp tables name 'temp_sv' and 'temp_sv_phenotype'.


jamie-m-a:
Looks fine

PR #783, opened 2022-01-17
Update ClinVar import (e107)

dglemos:
- Update version (ENSVAR-4439)
- Fix sql query to remove entries from allele_synonym and also remove entries from variation_citation (ENSVAR-4442)
- Update post process clinvar script (ENSVAR-4555)

### DIFF
--- a/scripts/import/import_clinvar_xml
+++ b/scripts/import/import_clinvar_xml
@@ -3,7 +3,7 @@
 =head1 LICENSE
 
 Copyright [1999-2015] Wellcome Trust Sanger Institute and the EMBL-European Bioinformatics Institute
-Copyright [2016-2021] EMBL-European Bioinformatics Institute
+Copyright [2016-2022] EMBL-European Bioinformatics Institute
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
--- a/scripts/import/import_clinvar_xml
+++ b/scripts/import/import_clinvar_xml
@@ -1449,13 +1449,13 @@ sub clean_old_data{
                                                     (select variation_id from variation v, source s
                                                     where v.source_id = s.source_id and s.name ='ClinVar')
                                                 ]);
-  $var_syn_sel_sth->execute();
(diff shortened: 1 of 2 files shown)


jamie-m-a (approved):

scripts/import/import_clinvar_xml
```diff
@@ -1441,7 +1443,23 @@ sub clean_old_data{
...
+  # first check which entries to remove
+  my $var_syn_sel_sth    = $dba->dbc()->prepare(qq[ select allele_synonym_id from allele_synonym where variation_id in
+                                                    (select variation_id from variation v, source s
+                                                    where v.source_id = s.source_id and s.name ='ClinVar')
+                                                ]);
+  $var_syn_sel_sth->execute();
+  my $data = $var_syn_sel_sth->fetchall_arrayref();
+  # remove entries from allele_synonym
+  foreach my $allele_syn (@{$data}){
+    my $allele_syn_id = $allele_syn->[0];
```
sarahhunt, line 1456: we don't need to refine the variable here

scripts/import/import_clinvar_xml
```diff
@@ -1441,7 +1443,23 @@ sub clean_old_data{
...
+    my $allele_syn_delete_sth = $dba->dbc()->prepare(qq[ delete from allele_synonym where allele_synonym_id = $allele_syn_id ]);
+    $allele_syn_delete_sth->execute();
```
sarahhunt, line 1458: It would be good to capture fails and die here, though the same goes for all the clean up steps.

sarahhunt, line 1458: It would be good to capture fails and die here, though the same goes for all the clean up steps.

sarahhunt (commented):
There are a couple of minor improvements possible here

scripts/import/import_clinvar_xml
dglemos: You mean we can use $allele_syn->[0] instead of $allele_syn_id?

sarahhunt:
merged to master & release/107

PR #784, opened 2022-01-17
Update citations import (e107)

dglemos:
ENSVAR-4556


helensch (commented):
Looks good for variants with multiple evidence_attribs
If the only evidence_attibs  for the variant is 'Cited' and this is removed with the SQL, it will set the evidence_attribs to the empy string and not the default NULL

jamie-m-a (approved):

PR #789, opened 2022-02-09
Updates to dbSNP for new cluster

helensch:
See: ENSVAR-4294


diegomscoelho (commented):
I've tested with and without `add_maf` new flag. Without add_maf it worked as expected. But when you include add_maf, you still have an error (also expected) in minor_allele > 50.

```sh
DBD::mysql::db do failed: Data too long for column 'minor_allele'
```

I would truncate minor_allele as it was done in hgvs, deleted_sequence and inserted_seq. Or just add '-' as it's done when frequency data is not present.

I would also include:

```sh
=item B<--no_maf>

Not include MAF information
```

https://github.com/Ensembl/ensembl-variation/pull/789/files#diff-1a080ddb11460f7cfa5030020d15d8a073021a6304bb18c2ec1a3b5fc6e991deR2281

diegomscoelho (approved):
LGTM! Commented alterations will be applied in a new PR.

PR #791, opened 2022-02-17
DB patch remove incorrect transcript_variation consequences

jamie-m-a:
Patch to remove the spurious consequences from transcript_variation table.

Tested on copy of dog DB and works.


dglemos (approved):

dglemos:
Merged into release/107 and master

PR #795, opened 2022-03-03
Update .travis.yml

olaaustine:
Ensembl-hive no longer has a master branch. 
Changing the travis.yml to reflect this so build do not fail.


dglemos:
@olaaustine thanks for the update, ensembl-test and ensemble should also be updated.

PR #800, opened 2022-04-04
Fix protein function pipeline for mouse - fetch refseq

dglemos:
Fetch only refseq transcripts
Details: ENSVAR-4481


olaaustine (approved):
Approved

PR #803, opened 2022-04-20
There should be no frequency when allele count is 0

nakib103:
When for some allele the allele count is 0 or undef we should set the frequency to be undef too. Currently it is set to AF (or 1 - AF depending on whether it is reference) which gives wrong information.

This happens only when reading from VCF. In the following figure China_south population do not have any T alleles but still get frequency of .078 from VCF file.

image


jamie-m-a:
Issue not specified correctly

PR #805, opened 2022-04-29
Dump GFF files from GO plugin

nuno-agostinho:
### [ENSVAR-4742](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-4742): VEP GO Plugin - web VEP + REST support

* GO files are only dumped for species for which there are variation databases
* _Monodelphis domestica_ is skipped for its large genome size (Tabix is limited to chromosomes with 512 Mbp or less)

GO files can be dumped as defined in the Confluence page: [Dump GO terms](https://www.ebi.ac.uk/seqdb/confluence/display/EV/Dump+GO+terms)

### DIFF
--- a/scripts/export/dump_GO_terms.pl
+++ b/scripts/export/dump_GO_terms.pl
@@ -90,29 +90,30 @@ $registry->load_registry_from_db(
 
 my %only_species = map { $_ => 1 } split(",",$species) if defined $species;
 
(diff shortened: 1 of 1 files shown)


dglemos (approved):

dglemos:
Merged into release/107 and master

PR #807, opened 2022-05-03
Renaming master to main

likhitha-surapaneni:
Renaming master to main for files dependent on `ensembl-variation` repo

For file, travisci/trigger-dependent-build.sh, there are some inconsistent lines as ensembl-rest is on main and ensembl-vep is on master. For now, keeping them the same and updating them after `ensembl-vep` is renamed to main


nuno-agostinho (approved):
Looks good to me, thanks @likhitha-surapaneni!

PR #810, opened 2022-05-12
Update transcript variation pipeline

dglemos:
For human: only write MANE transcripts to MTMP_transcript_variation
For other species: write all transcripts

ENSVAR-4767


jamie-m-a (approved):
Yep looks fine

PR #812, opened 2022-05-16
Fix ins instead of dup issue (hgvs)

nakib103:
For some HGVS input the hgvs output from VEP comes out as wrong - 
- the position comes out as wrong
- where it should be dup the output contains ins
see https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-2967 

**First,** for reverse stranded transcript, problem comes from hgvs parsing in `hgvs_from_component()`. When calculating genomic position for dup, strand is currently considered -

`forward stranded: start = end + 1`
-----------------s----------e--------------------> 
----------------------------es-------------------> 

`reverse stranded: start = end; end = end - 1`
<----------------e----------s--------------------- 
<---------------es-------------------------------- 

But, it need not be, as the position for a VariationFeature object always in the forward strand. For this reason reverse stranded transcript was causing issue.

**Second,** for `--hgvs_shift 0`, problem comes from how we match ref and alt allele string to see if that is dup when we form hgvs string in `hgvs_variant_notation()` function. We always look backward from end position of ref string, assuming the start position (before changed by `hgvs_from_component()` function with above logic) is at the left side of the end position. But for not reverse stranded transcript it should be at the right side of the end position -

----------------s+++++++e--------------------->
`start = end + 1`
-----------------+++++++es-------------------->
for `hgvs_transcript()` calculation is done with respect to the transcript
<-------------------------se+++++++------------

For shifting case though the start/end position move along the right side so checking the left side is alright.
----------------s+++++++e--------------------->
`start = end + 1`
-----------------+++++++es-------------------->
for `hgvs_transcript()` calculation is done with respect to the transcript
<-------------------------se+++++++------------
while shifted to 3` position the sequence we should check is still on left
<---------------------------+++++++se------------


likhitha-surapaneni:
Thank you for the PR @nakib103 . Tested the following things:
ENST00000299427.12:c.1548_1551dupTGAT : Works as expected
NM_001572.3:c.789_806dup : Works as expected
modules/t/hgvs_parser.t is failing for X:g.132081373dup and ENST00000298542.9:c.905+998dup for genomic and transcript levels: failed 4 tests of 663

nuno-agostinho:
Please check if this should go to `postreleasefix/108` instead of `main`.

likhitha-surapaneni (changes requested):
Hi, @nakib103. The tests may need to be fixed.

likhitha-surapaneni:
Hi @nakib103, I tried with some examples:

- ENST00000402135.3:c.1111_1114dup: the results seem to mostly match with the expected values. Location shows 2:58386932-58386933 and HGVSg=2:g.58386930_58386933dup, is that still right? I was expecting HGVSg=2:g.58386929_58386932dup (https://mutalyzer.nl/normalizer/GRCh37(chr2):g.58386929_58386932dup)
- ENST00000301067.7:c.11162_11167dup: I am wondering why we are not returning the 3' shifted notation for hgvsc: ENST00000301067.7:c.11168_11173dup (https://mutalyzer.nl/normalizer/ENST00000301067.7:c.11162_11167dup)

nakib103:
Hi @likhitha-surapaneni ,

- It is 1 bp shifted,  HGVSg=2:g.58386930_58386933dup looks alright. Mutalyzer also showing the same result - `NC_000002.11:g.58386930_58386933dup`
- did you run with `--shift_hgvs 0`? I got 3' shifted version -
```
$vep --id 'ENST00000301067.7:c.11168_11173dup' --database --db_version 108 --force --hgvs --shift_hgvs 1 --assembly GRCh37

#Uploaded_variation	Location	Allele	Gene	Feature	Feature_type	Consequence	cDNA_position	CDS_position	Protein_position	Amino_acids	Codons	Existing_variation	Extra
ENST00000301067.7:c.11168_11173dup	12:49427320-49427321	TGCAGC	ENSG00000167548	ENST00000301067	Transcript	inframe_insertion	11167-11168	11167-11168	3723	L/LQL	ctg/cTGCAGCtg	-	IMPACT=MODERATE;STRAND=-1;HGVSc=ENST00000301067.7:c.11168_11173dup;HGVSp=ENSP00000301067.7:p.Leu3723_Gln3724dup;HGVS_OFFSET=-6
```

likhitha-surapaneni (approved):
All the queries have been addressed. The PR works  for most of the cases.

likhitha-surapaneni:
Merged to main and release/110

PR #813, opened 2022-05-18
schema and test db pathde

nakib103:


modules/t/test-genome-DBs/homo_sapiens/funcgen/meta.txt
```diff
@@ -328,4 +328,6 @@
 767	\N	patch	patch_104_105_a.sql|schema_version
 768	\N	patch	patch_104_105_b.sql|New indices
 769	\N	patch	patch_105_106_a.sql|schema_version
-770     \N	patch   patch_106_107_a.sql|schema_version
+770	0
```
dglemos, line 331: You could delete this line.

PR #809, opened 2022-05-10
fix: --spdi flag quick fix

diegomscoelho:
## Description

As reported by issue https://github.com/Ensembl/ensembl-vep/issues/1120, flag `--spdi` is not working at all. Before we have more definitive solution, this PR enables output be created without synonyms.

## Test

1) Run vep with `--spdi` flag with and without this branch;

Old error:
```sh
Can't call method "db" on an undefined value at /<PATH-TO>/ensembl/modules/Bio/EnsEMBL/Slice.pm line 2984, <$fh> line 54.
```

This error should disappear and INFO field `SPDI` being populated, i.e: `21:6116502:C:T`

PR #822, opened 2022-06-15
fix: --spdi flag quick fix

diegomscoelho:
Jira: [Card](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-4727)

## Description

As reported by issue https://github.com/Ensembl/ensembl-vep/issues/1120, flag `--spdi` is not working at all. Before we have more definitive solution, this PR enables output be created without synonyms.

## Test

1) Run vep with `--spdi` flag with and without this branch;

Old error:
```sh
Can't call method "db" on an undefined value at /<PATH-TO>/ensembl/modules/Bio/EnsEMBL/Slice.pm line 2984, <$fh> line 54.
```

This error should disappear and INFO field `SPDI` being populated, i.e: `21:6116502:C:T`

PS.: #809 Reopening due to revert error

### DIFF
--- a/modules/Bio/EnsEMBL/Variation/VariationFeature.pm
+++ b/modules/Bio/EnsEMBL/Variation/VariationFeature.pm
@@ -2070,7 +2070,9 @@ sub spdi_genomic{
   my $vf_strand = $self->strand();
 
   # set up sequence reference
-  my $syn = $ref_slice->get_all_synonyms('RefSeq_genomic');
+  # Quick fix to spdi flag works, needs to be investigated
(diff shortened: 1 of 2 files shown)


likhitha-surapaneni (approved):
Hi @diegomscoelho , thank you for the fix.
Tested VEP with `--spdi ` flag with and without the PR and the issue seems to be only for `--offline` mode. 
This PR resolves the issue and the results are the same for `offline`, `cache` and `database `options.
The changes look good to me

nuno-agostinho:
Merged into release/108 and main. Cherry-picked to release/107.

PR #823, opened 2022-06-17
fix: spdi --flag keep get_all_synonyms()

diegomscoelho:
## Description

It seems that we need to keep get_all_synonyms() in case adaptor->db still exists.

## Test

1) Run VEP with --spdi flag


nuno-agostinho (approved):
Looks good to me!

nuno-agostinho:
Merged into release/108 and main. Cherry-picked to release/107.

PR #824, opened 2022-06-17
Update dump phenotypes to include boxer

dglemos:


jamie-m-a (approved):
Regex working as expected, looks good to me.

PR #827, opened 2022-06-21
Fix species list doc creation script

nakib103:
Currently we have species database which name can contain numbers such as - 
**Gallus_gallus_gca000002315v5**

Updated regex to handle that. Also in the species name we do not want the `gca` to appear.

### DIFF
--- a/scripts/misc/species_list.pl
+++ b/scripts/misc/species_list.pl
@@ -146,7 +146,7 @@ foreach my $hostname (@hostnames) {
     $species_list{$s_name}{'count'} = round_count($count_var);
 
     my $label_name = ucfirst($s_name);
-       $label_name =~ s/gca[0-9v]+//g;	# remove any gca from name 
+       $label_name =~ s/_gca[0-9]{9}[v0-9]*+$//g;	# remove any gca from name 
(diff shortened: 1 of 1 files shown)


scripts/misc/species_list.pl
```diff
@@ -146,6 +146,7 @@ =head1 CONTACT
     $species_list{$s_name}{'count'} = round_count($count_var);
 
     my $label_name = ucfirst($s_name);
+       $label_name =~ s/gca[0-9v]+//g;	# remove any gca from name
```
nuno-agostinho, line 149: NB: This would impact any species that contains "gcav" in its scientific name. We don't have any such species in Ensembl (and from a quick Google search, there doesn't seem to be any such species.

Still, I would suggest the following to ensure we only strip out the proper GCA from the species name.

```suggestion
       $label_name =~ s/_gca[0-9]{9}[v0-9]*+$//g;	# remove any gca from name 
```

nuno-agostinho (commented):
Thanks @nakib103! Seems fine, I only have one comment.

nuno-agostinho (approved):

PR #829, opened 2022-06-24
Update import_vcf.pl

dglemos:


scripts/import/import_vcf.pl
```diff
@@ -745,6 +745,8 @@ sub main {
       }
 
       my $chromosome = $data->{tmp_vf}->{chr};
+      $chromosome =~ s/Chr//i if(defined $config->{use_chr});
```
diegomscoelho, line 748: It's not better to remove by default this prefix in any case? - If so, I would remove this flag `use_chr` also.

```suggestion
      $chromosome =~ s/Chr//i;
```

nuno-agostinho: I agree with Diego and I would say to extend it to also delete the word `Chromosome` at it appears in [GCA_000298735.2 (sheep, texel)](http://ftp.ebi.ac.uk/pub/databases/eva/rs_releases/release_3/by_species/ovis_aries/GCA_000298735.2):
```suggestion
      $chromosome =~ s/chr(omosome)?//i;
```

olaaustine: We can not make it default as this scripts is used for other imports that are not EVA and also used by other teams.
Thats why it can not be used as a default and can only be used as an option.

diegomscoelho: LOTM!

olaaustine:
Merged into release/108 and main

PR #830, opened 2022-06-27
Add a boolean flag in VCF config to fetch consequences from VCF file

likhitha-surapaneni:
[ENSVAR-4961](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-4961)

This PR attempts to add a boolean flag to VCF config file that allows fetching consequences from VCF file instead of calculating on fly

This flag is then used to set dont_fetch_vf_overlaps in get_all_VariationFeatures_by_Slice in https://github.com/Ensembl/ensembl-webcode/blob/78c0e72a29ee87ea9e2ee9f2981a95cccb49cb34/modules/EnsEMBL/Web/Query/GlyphSet/Variation.pm#L236

### DIFF
--- a/modules/Bio/EnsEMBL/Variation/BaseAnnotation.pm
+++ b/modules/Bio/EnsEMBL/Variation/BaseAnnotation.pm
@@ -81,6 +81,7 @@ sub new {
     $adaptor,
     $use_seq_region_synonyms,
     $tmpdir,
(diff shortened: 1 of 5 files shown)


nakib103 (approved):
Tested with API too. Seems to work fine. LOTM!

PR #834, opened 2022-06-29
Update to make unique ids among dog breeds

nakib103:
Dog and Dog boxer had same id which cause problem with the API.


dglemos:
Merged into release/108, release/107 and main

PR #837, opened 2022-06-30
Changes to skip based on length check

olaaustine:


likhitha-surapaneni:
Changes look good to me. Tested and works as expected

PR #844, opened 2022-07-19
fix: Rebuild indexes not working in normal mode

diegomscoelho:
## Description

When running full mode, I noticed that rebuilding indexes was not working properly. This PR fix workflow logic to run properly index rebuild after running web_index_load

## Test

1) Just need to check that rebuild indexes is triggered by init_transcript_effect as previsouly.


jamie-m-a (changes requested):
the -wait_for is not always reliable and can start sooner than you expect depending on when the job is created. It's usually safer to use a semaphore structure when an analysis is dependent on a previous one: https://ensembl-hive.readthedocs.io/en/version-2.6/quickstart/tips.html#using-semaphore-funnels

jamie-m-a (approved):
Looks fine - I'll apply the fix locally and test on my run then merge later if successful.

PR #854, opened 2022-07-26
Merge pull request #838 from nakib103/vcf_many_synonym

nakib103:
This PR is needed for RR37 as that release uses e107 branches. 
We are releasing pike parch which needs this fix to show the tracks.

### DIFF
--- a/modules/Bio/EnsEMBL/Variation/VCFCollection.pm
+++ b/modules/Bio/EnsEMBL/Variation/VCFCollection.pm
@@ -1291,9 +1291,21 @@ sub _get_vcf_by_chr {
   if(!exists($self->{files}) || !exists($self->{files}->{$chr})) {
     my $obj;
     
-    # check we have this chromosome
+    # check we have this chromosome or its synonym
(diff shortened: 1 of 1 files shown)


nuno-agostinho (approved):
Looks good to me!

Just two suggestions for the future: could the prefix be case-insensitive (some species have `ChrX` for instance) or even customised with an option called `prefix` or similar in the JSON config?

nakib103:
> Looks good to me!
> 
> Just two suggestions for the future: could the prefix be case-insensitive (some species have `ChrX` for instance) or even customised with an option called `prefix` or similar in the JSON config?

The thing is the vcf file is parsed by an parser from `ensembl-io` and it only supports `chr` prefix currently. So even we add support for case-insensitive or other prefix here it will not work. But if support for those prefixes are added to `ensembl-io` we can and should also add support for those here.

nakib103:
This PR will be needed to merge only if for an e107 equivalent RR release we have a species that have non-chromosomal seq region that are too many to put in the vcf config file.

This fix is already merged to 108 branch.

PR #856, opened 2022-07-28
Fix ClinVar import

dglemos:
This PR is just to update the ClinVar script, COSMIC will be updated later.
Details in: ENSVAR-5014

### DIFF
--- a/scripts/import/import_clinvar_xml
+++ b/scripts/import/import_clinvar_xml
@@ -1337,10 +1337,13 @@ sub enter_var{
 
   $var_feat_adaptor->store($vf);
 
+  my $count_tv = 0;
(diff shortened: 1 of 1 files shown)


nuno-agostinho (approved):
Thanks @dglemos! Tested running with the example file in ENSVAR-5014, seems to be working as intended:
- Before, variant `rs2046013117` was added to `variation_feature` with consequences `downstream_gene_variant,upstream_gene_variant,regulatory_region_variant,TF_binding_site_variant`.
- Thanks to this update, the consequence of `rs2046013117` is set to the default `intergenic_variant`, as expected.

nuno-agostinho:
Merged to main and release/108.

PR #861, opened 2022-08-08
Removing variants that break CAR API

jamie-m-a:
As per Jira ticket - ENSVAR-4991.

Tested command and generates output files correctly.

### DIFF
--- a/modules/Bio/EnsEMBL/Variation/Pipeline/GetCAR/GetCARSeqname.pm
+++ b/modules/Bio/EnsEMBL/Variation/Pipeline/GetCAR/GetCARSeqname.pm
@@ -99,7 +99,7 @@ sub run {
     my $json_file = $filename . '-car-lu' . '.json';
 
     # Getting the HGVS value from the full file
-    my $cmd = "cut -f6 $hgvs_dir/$hgvs_file | grep -v ] > $car_lu_dir/$lu_file";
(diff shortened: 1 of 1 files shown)


modules/Bio/EnsEMBL/Variation/Pipeline/GetCAR/GetCARSeqname.pm
```diff
@@ -99,7 +99,7 @@ sub run {
     my $json_file = $filename . '-car-lu' . '.json';
 
     # Getting the HGVS value from the full file
-    my $cmd = "cut -f6 $hgvs_dir/$hgvs_file > $car_lu_dir/$lu_file";
+    my $cmd = "cut -f6 $hgvs_dir/$hgvs_file | grep -v ] > $car_lu_dir/$lu_file";
```
diegomscoelho, line 102: To avoid skipping non redundant ID, I think it could be better skipping just ending with `]`. What you think?
```suggestion
    my $cmd = "cut -f6 $hgvs_dir/$hgvs_file | grep -v ]$ > $car_lu_dir/$lu_file";
```

diegomscoelho (commented):
Small suggestion

jamie-m-a:
Agree, I don't think it's possible to get an ID like that, but just in case!

diegomscoelho (approved):
LGTM!

PR #862, opened 2022-08-09
fix: Delete from GeneFactory in smaller chunks

diegomscoelho:
Jira: [Card](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-5019)

## Description

As reported by Jamie, we had a few transcripts that were not deleted during update mode for VariationConsequence Pipeline. This PR divides `DELETE FROM` command in smaller chunks similar to UpdateVariationFeature task to avoid any long string problem.

## Test

1) Running VariationConsequence Pipeline and checking after `GeneFactory` if all deleted transcripts were removed.


jamie-m-a (approved):

jamie-m-a:
All transcripts correctly removed now - thanks!

PR #866, opened 2022-08-12
Restoring config for parus_major to use the original VCF

likhitha-surapaneni:
The study we used to point to at the EVA contains a load of genotyping info that we’ll lose by pointing to the new VCF.

This PR restores the config to point to the original VCF


jamie-m-a (approved):
Looks good - thanks!

PR #870, opened 2022-08-17
Add SO term complex_chromosomal_rearrangement

dglemos:


nakib103 (commented):
@dglemos , I think we should also update the `VARIATION CLASS` at https://github.com/Ensembl/ensembl-variation/blob/8a615767ef196e7a7fb06daa3d9f390393240bb8/modules/Bio/EnsEMBL/Variation/Utils/Constants.pm#L132.

Because the doc generating script use that variable too - https://github.com/Ensembl/ensembl-variation/blob/8a615767ef196e7a7fb06daa3d9f390393240bb8/scripts/misc/generate_classes_table.pl#L198

dglemos:
@nakib103 , good spot! It's added.

nakib103 (approved):
LOTM! ~

PR #876, opened 2022-08-25
Schema and test db patches for e109

nakib103:


modules/t/test-genome-DBs/homo_sapiens/core/external_db.txt
```diff
@@ -343,3 +343,4 @@
 230	Ribosomal_Protein_Gene_Data	Dec.2010	XREF	1	0		Ribosomal Protein Gene DB	MISC	\N
 50612	LampreyGenomeProject		KNOWN	0	0	ARRAY	Lamprey Genome Project	MISC	\N
 50620	Illumina_Bodymap_transcript		XREF	0	0		Illumina Bodymap transcriptome	MISC	\N
+0	GeneCards		XREF	50	GeneCards	MISC	\N	\N	\N
```
dglemos, line 346: Can you check why the id is zero?

dglemos (commented):
The file [table.sql](https://github.com/Ensembl/ensembl-variation/blob/20d61b0a0b03e499c066175524c44214b2b7093f/sql/table.sql#L1838) also needs to be updated with the new schema version and the new patch.

dglemos:
Tests failing are not caused by this PR.

PR #882, opened 2022-09-08
VCFCollection - extract CLNSIG info

Gurpreet-Ghattaoraya:
Using "$vf->get_all_clinical_significance_states()"

### DIFF
--- a/modules/Bio/EnsEMBL/Variation/VCFCollection.pm
+++ b/modules/Bio/EnsEMBL/Variation/VCFCollection.pm
@@ -486,59 +486,77 @@ sub get_all_VariationFeatures_by_Slice {
         $desc = $val->{Description};
         $vcf_info_field = 'ANN';
         last;
+      }
+      elsif($val->{ID} eq 'CLNSIG'){
+        $desc = $val->{Description};
+        $vcf_info_field = 'CLNSIG';
+        last;
       } 
     }
-    my @description = split(/Format:/,$desc);
-    my @info_format = split('\|', $description[1]);
-    foreach my $vf (@vfs) { 
-      if ( $vcf_info_field ){
-        my $info = $vf->{vcf_record}->get_info;
-        my @vcf_variants = $info->{$vcf_info_field};
-        foreach my $vcf_variant ( @vcf_variants ) {
-          if ( defined($vcf_variant) ){
-            my @transcript_vars = split(',',$vcf_variant);
-            foreach my $transcript_var ( @transcript_vars ){
-              my @transcript_split = split('\|',$transcript_var);
-              my %transcript_map = zip @info_format, @transcript_split;
-              next if not defined($transcript_map{'Feature_type'});
-              if ( $transcript_map{'Feature_type'} eq 'Transcript') {
-                my $tv = Bio::EnsEMBL::Variation::TranscriptVariation->new_fast({
-                  variation_feature  => $vf,
-                  _feature_stable_id => $transcript_map{'Feature'}
-                });  
-                #Add variation Feature seq from transcript map
-                my $tva = Bio::EnsEMBL::Variation::TranscriptVariationAllele->new_fast({
-                  variation_feature_seq => $transcript_split[0],
-                  is_reference               => 0
-                });
-                my @cons_list = split('&',$transcript_map{'Consequence'});
-                foreach my $con (@cons_list){
-                  if ( exists $OVERLAP_CONSEQUENCES{$con} ) {
-                    my $new_cons = $OVERLAP_CONSEQUENCES{$con};
-                    $tva->add_OverlapConsequence($new_cons);
+    if ($vcf_info_field eq 'CSQ' or $vcf_info_field eq 'ANN'){
+      my @description = split(/Format:/,$desc);
+      my @info_format = split('\|', $description[1]);
+      foreach my $vf (@vfs) { 
+        if ( $vcf_info_field ){
+          my $info = $vf->{vcf_record}->get_info;
+          my @vcf_variants = $info->{$vcf_info_field};
+          foreach my $vcf_variant ( @vcf_variants ) {
+            if ( defined($vcf_variant) ){
+              my @transcript_vars = split(',',$vcf_variant);
+              foreach my $transcript_var ( @transcript_vars ){
+                my @transcript_split = split('\|',$transcript_var);
+                my %transcript_map = zip @info_format, @transcript_split;
+                next if not defined($transcript_map{'Feature_type'});
+                if ( $transcript_map{'Feature_type'} eq 'Transcript') {
+                  my $tv = Bio::EnsEMBL::Variation::TranscriptVariation->new_fast({
+                    variation_feature  => $vf,
+                    _feature_stable_id => $transcript_map{'Feature'}
+                  });  
+                  #Add variation Feature seq from transcript map
+                  my $tva = Bio::EnsEMBL::Variation::TranscriptVariationAllele->new_fast({
+                    variation_feature_seq => $transcript_split[0],
+                    is_reference               => 0
+                  });
+                  my @cons_list = split('&',$transcript_map{'Consequence'});
+                  foreach my $con (@cons_list){
+                    if ( exists $OVERLAP_CONSEQUENCES{$con} ) {
+                      my $new_cons = $OVERLAP_CONSEQUENCES{$con};
+                      $tva->add_OverlapConsequence($new_cons);
+                    }
+                    else{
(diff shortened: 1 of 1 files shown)


modules/Bio/EnsEMBL/Variation/VCFCollection.pm
```diff
@@ -486,59 +486,77 @@ sub get_all_VariationFeatures_by_Slice {
...
-          -no_ref_check       => 1,
-        );
-        weaken($vf->{intergenic_variation}->{base_variation_feature});
+    } elsif ($vcf_info_field eq 'CLNSIG'){
+      foreach my $vf (@vfs) {
+        my $info = $vf->{vcf_record}->get_info;
+        my $vcf_variants = $info->{$vcf_info_field};
+        my @clnsig_vars = split('\|',$vcf_variants);
+        foreach my $clnsig_var (@clnsig_vars) {
+          $clnsig_var =~ s/_/ /g;
```
nakib103, line 552: According to JIRA - https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-5046 we have some specific format for ClinVar clinical significance. We also put them out in our doc like this - https://www.ensembl.org/info/genome/variation/phenotype/phenotype_annotation.html#clin_significance . So, we should add the clinical significance term according to this list.

For example, It would be better to have some more filtering. Like -
```
# removing leading underscore _
_risk_factor -> risk factor
# splitting on / and further processing
Pathogenic/Likely_pathogenic -> we will get pathogenic and likely pathogenic
```
I found these two cases from the values from chromosome 1 of the ClinVar VCF. You may find some more.

And, maybe you should filter out the value also so that it matches exactly with our attribs here - https://github.com/Ensembl/ensembl-variation/blob/main/modules/Bio/EnsEMBL/Variation/Utils/Config.pm#L68
You can import the array in this file and do a grep before adding them to the `$vf->{clinical_significance}`

nakib103 (changes requested):

modules/Bio/EnsEMBL/Variation/VCFCollection.pm
```diff
@@ -486,59 +486,77 @@ sub get_all_VariationFeatures_by_Slice {
...
-        $vf->{intergenic_variation} = Bio::EnsEMBL::Variation::IntergenicVariation->new(
-          -variation_feature  => $vf,
-          -no_ref_check       => 1,
-        );
-        weaken($vf->{intergenic_variation}->{base_variation_feature});
+    } elsif ($vcf_info_field eq 'CLNSIG'){
+      foreach my $vf (@vfs) {
+        my $info = $vf->{vcf_record}->get_info;
+        my $vcf_variants = $info->{$vcf_info_field};
+        my @clnsig_vars = split('\|',$vcf_variants);
```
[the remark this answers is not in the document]
Gurpreet-Ghattaoraya: The VCF file I've been working with described in its header that it was using "|" to separate.  I looked in the file to see this was the case. I shall see if I can use multiple delimiters to split by.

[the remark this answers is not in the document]
Gurpreet-Ghattaoraya: Speaking with @nakib103 - we agreed using `split('[\|,]')` to split on either delimeter. (needs testing)

modules/Bio/EnsEMBL/Variation/VCFCollection.pm
Gurpreet-Ghattaoraya: I'm adding line to trim leading and trailing whitespaces from the string: `$clnsig_lc =~ s/^\s+|\s+$//;`.
For the splitting on "/" I have included the forward slash as another character by which to split by - since the split happens before any processing takes place. I've tested and the example "benign/likely_benign" will come out as two entries of "benign" and "likely benign"

modules/Bio/EnsEMBL/Variation/VCFCollection.pm
```diff
@@ -486,59 +486,77 @@ sub get_all_VariationFeatures_by_Slice {
...
+        foreach my $clnsig_var (@clnsig_vars) {
+          $clnsig_var =~ s/_/ /g;
+          my $clnsig_lc = lc($clnsig_var);
+          my $clnsig = Bio::EnsEMBL::Variation::VariationFeature->new_fast({clinical_significance => $clnsig_lc });
+          push @{ $vf->{clinical_significance} ||= [] }, $clnsig;
+        };
```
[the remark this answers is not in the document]
Gurpreet-Ghattaoraya: I'm confused here, my IDE says this is the closing bracket for `foreach my $clnsig_var (@clnsig_vars) {`
But I was struggling with balancing my brackets so I will check agian. with @nakib103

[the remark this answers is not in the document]
Gurpreet-Ghattaoraya: Speaking with @nakib103 - I now understand what he meant.

nakib103 (approved):
Looks good! ~

modules/t/VCFCollection.t
```diff
@@ -488,4 +488,21 @@ is_deeply(
...
+my $temp = $coll->filename_template();
+$temp =~ s/###t\-root###/$dir/;
+$coll->filename_template($temp);
+ok($coll->filename_template =~ /^$dir/, "update filename_template");
+
+$slice = $sa->fetch_by_region('toplevel', 1, 10, 20);
+my $dont_fetch_vf_overlaps=1;
+my $vfs = $coll->get_all_VariationFeatures_by_Slice($slice, $dont_fetch_vf_overlaps);
+
+ok($vfs->[0]->get_all_clinical_significance_states()->[0] eq 'likely benign', 'get_all_clinical_significance_states - obtain single clinical significance entry');
```
dglemos, line 504: The test fails because the correct clinical significance is `uncertain significance`.

modules/t/VCFCollection.t
```diff
@@ -488,4 +488,21 @@ is_deeply(
...
+ok(scalar (@{$vfs->[1]->get_all_clinical_significance_states()}) eq 1, 'get_all_clinical_significance_states - ignore upsupported clinical significance entry');
+ok(scalar (@{$vfs->[2]->get_all_clinical_significance_states()}) eq 2, 'get_all_clinical_significance_states - obtain multiple clinical significance entries');
```
dglemos, line 506: `$vfs->[2]` only has one clinical significance -> 'likely benign'

Gurpreet-Ghattaoraya: Yes, I was using an old version of the VCF file in this test. I have commited a new version of the VCF file which should hopefully fix.

dglemos:
The file has to be test_vcf_clnsig.vcf.gz

modules/t/VCFCollection.t
dglemos: You loaded a new vcf file but kept the old files.

dglemos: One of the tests is still going to fail: `ok(scalar (@{$vfs->[2]->get_all_clinical_significance_states()}) eq 2` only has one valid clinical significance.

Gurpreet-Ghattaoraya: Apologies, I comitted the wrong file. I used the .vcf file and not the .vcf.gz

modules/Bio/EnsEMBL/Variation/VCFCollection.pm
```diff
@@ -486,59 +487,79 @@ sub get_all_VariationFeatures_by_Slice {
...
-        $vf->{intergenic_variation} = Bio::EnsEMBL::Variation::IntergenicVariation->new(
-          -variation_feature  => $vf,
-          -no_ref_check       => 1,
-        );
-        weaken($vf->{intergenic_variation}->{base_variation_feature});
+    } elsif ($vcf_info_field eq 'CLNSIG'){
+      foreach my $vf (@vfs) {
+        my $info = $vf->{vcf_record}->get_info;
+        my $vcf_variants = $info->{$vcf_info_field};
+        my @clnsig_vars = split('[\|,/]',$vcf_variants);
```
dglemos, line 551: As there are clinsig terms with commas, you can deal with the two specific terms here.
```
my $vcf_variants = $info->{$vcf_info_field};
# replace commas for specific examples
$vcf_variants =~ s/Pathogenic,_low_penetrance/Pathogenic_low_penetrance/;
$vcf_variants =~ s/Likely_pathogenic,_low_penetrance/Likely_pathogenic_low_penetrance/;
my @clnsig_vars = split('[\|,/]',$vcf_variants);
```

PR #885, opened 2022-09-14
Update SV count in the doc

nakib103:
link - http://wp-np2-11.ebi.ac.uk:7070/info/genome/variation/species/species_detailed_counts.html 

We were showing SV count from only DGVa source. Now we have added dbVar too.


jamie-m-a (approved):
Looks good!

PR #887, opened 2022-09-16
Update simulate_variation script to generate SpliceAI files

dglemos:
Update the script to generate SpliceAI files (-spliceai)
Add option to generate variants only overlapping MANE transcripts (-onlyMane)

### DIFF
--- a/tools/variant_simulator/simulate_variation
+++ b/tools/variant_simulator/simulate_variation
@@ -60,7 +60,7 @@ usage() if ($help);
 
 die "Illegal arguments -exonsOnly -intronsOnly can't be used together !" if $exonOnly && $intronOnly;
 die "Illegal arguments -intronsOnly -codingsOnly can't be used together !" if $codingOnly && $intronOnly;
-die "Illegal arguments -onlyMane -codingsOnly -exonsOnly can't be used together !" if $onlyMane && $exonOnly && $codingOnly;
+die "Illegal arguments -onlyMane can only be used alone !" if $onlyMane && ($exonOnly || $codingOnly || $intronOnly);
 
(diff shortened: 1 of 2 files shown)


tools/variant_simulator/simulate_variation
```diff
@@ -34,30 +34,33 @@ use Bio::EnsEMBL::Registry;
...
+    "spliceai"    => \$spliceai,
+    "registry=s"  => \$registryFile,
     "help!" => \$help,
 ) or usage();
 
 usage() if ($help);
 
 die "Illegal arguments -exonsOnly -intronsOnly can't be used together !" if $exonOnly && $intronOnly;
 die "Illegal arguments -intronsOnly -codingsOnly can't be used together !" if $codingOnly && $intronOnly;
+die "Illegal arguments -spliceai -intronsOnly -codingsOnly -exonsOnly can't be used together !" if $spliceai && $intronOnly && $exonOnly && $codingOnly;
```
diegomscoelho, line 63: This will only work if all four options are used at same time, which already happen at previous die. All those four options should be triggered together, or any pairwise from those four should not be triggered?

diegomscoelho (commented):
LOTM! Both commands are working, have a small suggestion in `die` part.

tools/variant_simulator/simulate_variation
dglemos: I fixed the condition. -spliceai can be used with the other ones, only mane can't.

diegomscoelho: Thanks! Solved.

diegomscoelho (approved):
LGTM!

PR #889, opened 2022-09-21
Add new attrib_type for CGC phenotypes

dglemos:
We import more data from CGC, we need to add 3 attrib_types to store the extra data: [ENSVAR-3591](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-3591)


olaaustine (approved):

olaaustine:
Merged into main and release/109

PR #891, opened 2022-09-27
Update script import_variant_synonyms

dglemos:


nakib103 (approved):
variation_set_id can be empty string in `variation_feature` table. 
Looks good! ~

PR #893, opened 2022-09-28
Update SV import 

dglemos:
The newest ClinVar file has encoded clinical significance:
`clinical_int=Likely%20pathogenic`


nuno-agostinho (approved):
Thanks for the update @dglemos!

Everything seems to run as expected:
- All clinical significance results are decoded in `structural_variation` from variation DB
- Internal database now tracks studies imported into variation DB:
  - `check_result` contains imported studies
  - `ensvar_db` contains variation DB modified by the import script
  -  `ensvardb_status` states run type for variation DB

PR #895, opened 2022-09-30
Add attrib for DDG2P citations

dglemos:
Add new attrib for release 110
[ENSVAR-4990](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-4990)


diegomscoelho (approved):
LOTM! Tested and does it work.

PR #898, opened 2022-10-05
Restrucuring unittest for fetching consequences from VCF

likhitha-surapaneni:
Removed `vcf_consequences.t` and merged the code to `VCFCollections.t `


nakib103 (approved):
Looks Good!~
Coverage decreasing is from removal of common code which is now redundant.

PR #903, opened 2022-10-14
New script to check which SV studies to update

dglemos:
[ENSVAR-5204](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-5204)

### DIFF
--- a/scripts/python/check_sv_updates.py
+++ b/scripts/python/check_sv_updates.py
@@ -1,5 +1,5 @@
 """
-Version: 1.0 (2022-08-22)
+Version: 1.0 (2022-10-14)
 """
 
 import argparse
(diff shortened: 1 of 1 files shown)


scripts/python/check_sv_updates.py
```diff
@@ -0,0 +1,160 @@
+"""
```
nuno-agostinho, line 1: Add python3 shebang:

```suggestion
#!/usr/bin/env python3                                                          

"""
```

nuno-agostinho (approved):
Hey @dglemos, could you add a python3 shebang to the script?

The script seems to return the expected result when running:
```
./check_sv_updates.py -r 109 --user ensro --host mysql-ens-var-prod-3 --port 4606 --prod-host mysql-ens-var-prod-1 --prod-port 4449 --report-all
```

PR #904, opened 2022-10-17
Fix CAR import pipeline

nakib103:


jamie-m-a (changes requested):
So this seems fine and does work, but maybe we can consider filtering out at an earlier stage to avoid having to mess around with counters within counters.

So at https://github.com/Ensembl/ensembl-variation/blob/a885ada3058bcd8f284c09396e7d911341f8aec2/modules/Bio/EnsEMBL/Variation/Pipeline/DumpHGVS/DumpRegionHGVS.pm#L94 

We could filter the same way GetCARSeqname.pm, that way the input file will match and there won't be the potential for mismatches later?

nakib103:
Thanks @jamie-m-a , I think that makes sense and better approach. I have updated accordingly.

PR #906, opened 2022-10-20
Fix bug in SV import (internal db)

dglemos:
Use study name when updating column `is_current` in the internal db (production db)


nuno-agostinho (changes requested):

nuno-agostinho (approved):

nuno-agostinho:
Tested the new changes and everything seems to work out as expected:
* The results now show only file name without path and without extension: `path/nstd102.gvf` -> `nstd102`
* `is_current` now shows 1 to all current databases, even we run the script twice (one to update old data and another to import new data)

Thanks, @dglemos!

nuno-agostinho:
Merged to `release/109` and `main`.

PR #907, opened 2022-10-25
Prevent gvf file corruption in Release dump pipeline

nakib103:
If the release dump pipeline is stopped in the middle of gvf file dumping from DB and restarted again, it starts appending on the already existing file. It results in gvf file corruption.

To prevent this, check if file already there and is not empty and just delete the file before starting gvf dump all over again.


jamie-m-a (approved):
Looks good to me.

PR #912, opened 2022-11-17
fix: Do not run if update_diff is wrong

diegomscoelho:
Jira: [Card](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-5255)

## Description

While running update_mode, if you set a wrong `--update_diff` filename pipeline understands that you have started a full mode wrongly. This PR creates a `die` to crash in first task to advise that user need to introduce correct name.

## Test

1) Start a VariationConsequence pipeline with `--update_diff` followed by a wrong name.
2) It should get: `File used in flag --update_diff <PATH-TO>/<FILE> does not exist`


likhitha-surapaneni (approved):
Looks good to me

PR #916, opened 2022-11-22
Remove outdated citations from phenotype import

nuno-agostinho:
[ENSVAR-5011](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-5011)

After phenotype import, if citations are not available in `phenotype_feature` or `phenotype_feature_attrib`:
- remove citation if outdated in all sources
- if only some sources are outdated, discard only those sources

You can test the code on my database: `nuno_homo_sapiens_variation_109_38_outdated_citations` in `v2`.


scripts/import/import_EPMC.pl
```diff
@@ -1265,8 +1278,23 @@ sub check_outdated_citations {
...
+    
+    if (!@current_attribs) {
+      # remove citation if outdated in all sources
+      $rm_citations_sth->execute($attrib_id, $variation_id, $publication_id) ||
+        die "Error: cannot remove $variation_id, $publication_id, $attrib_id from variation_citation\n";
+    } elsif (@current_attribs ne @split_attrib_id) {
+      # discard outdated sources
+      my $new_attrib_id = join(",", @current_attribs);
+      $update_citations_sth->execute($new_attrib_id, $variation_id, $publication_id) ||
+        die "Error: cannot update $variation_id, $publication_id, $attrib_id from variation_citation with new value $new_attrib_id\n";
```
nuno-agostinho, line 1297: Should we use `$variation_id` and `$publication_id` (more direct way to see where the problem is) or instead `$variaiton_rsid` and `$publication_pmid` as used for the output file?

scripts/import/import_EPMC.pl
```diff
@@ -1265,8 +1278,23 @@ sub check_outdated_citations {
 
     my @split_attrib_id = split /,/, $attrib_id;
 
+    my @current_attribs;
     foreach my $attrib (@split_attrib_id) {
-      print $wrt "$variation_rsid\t$publication_pmid\t$attrib\n" unless ($citations_pheno_feature->{$attrib.'_'.$publication_pmid} || $citations_pheno_feature_attrib->{$attrib.'_'.$publication_pmid});
+      unless ($citations_pheno_feature->{$attrib.'_'.$publication_pmid} || $citations_pheno_feature_attrib->{$attrib.'_'.$publication_pmid}) {
+        push @current_attribs, $attrib;
+        print $wrt "$variation_rsid\t$publication_pmid\t$attrib\n";
```
nuno-agostinho, line 1285: Is the file still useful? I can remove this part if not.

nuno-agostinho:
@dglemos Makes sense, updated to go against main, thanks!

dglemos (dismissed):
The method remove_outdated_citations() is removing the citations that were just inserted from the phenotype table.

scripts/import/import_EPMC.pl
```diff
@@ -1265,8 +1278,23 @@ sub check_outdated_citations {
...
+      }
+    }
+    
+    if (!@current_attribs) {
```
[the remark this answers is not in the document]
nuno-agostinho: Yes, you are right. At first, `@current_attribs` should hold the updated attribs. I updated its name to `@outdated_attribs` and changed the logic.

dglemos:
@nuno-agostinho, thanks for the update! The script is not deleting the newly inserted citations anymore but it's deleting more entries than it should. An example is:
In table variation_citation, a variant has `data_source_attrib = 615,620` and the citation is outdated then it should only delete the attrib `620` because we don't know if it's still being cited from EPMC (615). However, the script is deleting the entry from the table. 
After the import, it should be: `data_source_attrib = 615` but instead the entry is removed.

On another note, there are a few extra checks the script should do. When an entry is deleted from variation_citation we have to check if the variants is still being cited for other publications/sources, if the variant is not cited anymore then we have to update the column evidence_attribs in tables variation and variation_citation. It's a similar logic to this method: https://github.com/Ensembl/ensembl-variation/blob/main/scripts/import/import_EPMC.pl#L914

scripts/import/import_EPMC.pl
```diff
@@ -1282,11 +1242,49 @@ sub check_outdated_citations {
...
+    where publication_id NOT IN
+      (select distinct publication_id from variation_citation) ]);
+  $orphan_publications_sth->execute() or die;
+  my $orphan_publications = $orphan_publications_sth->fetchall_arrayref();
+
+  if ($orphan_publications) {
+    print $wrt "\nDeleted publications with no associated variants (publication_id, title):\n";
+    remove_publications($pub_ad, $wrt, $orphan_publications);
+  } else {
+    print $wrt "\nNo publications with no associated variants to remove.\n";
```
dglemos, line 1285: The evidence_attribs is only being updated by `remove_publications()`. If no publications are removed then the evidence_attribs is never updated, this is a problem when a variant doesn't have any more citations.
An alternative is to update [update_evidence()](https://github.com/Ensembl/ensembl-variation/blob/59efd6ddd9e3c77019f45624852248fec8905cf5/scripts/import/import_EPMC.pl#L595), this method could add/remove cited evidence instead of just adding it.

dglemos (changes requested):
Thanks for the updates.
The update works well - I only found one issue:
when a citation is outdated after it has been removed and there are no more citations for that variant the `variation.evidence_attribs` still has the evidence `371 (Cited)`. I think the problem is in [line 1281](https://github.com/Ensembl/ensembl-variation/blob/59efd6ddd9e3c77019f45624852248fec8905cf5/scripts/import/import_EPMC.pl#L1281-L1286)

scripts/import/import_EPMC.pl
```diff
@@ -132,7 +132,8 @@
   my $citations_pheno_feature_attrib = process_phenotype_feature_attrib($dba, $source_ad, $citation_attribs);
   import_citations($reg, $citations_pheno_feature_attrib, $type);
 
-  check_outdated_citations($dba, $var_ad, $pub_ad, $citation_attribs, $citations_pheno_feature, $citations_pheno_feature_attrib);
+  print STDERR "Removing outdated citations...\n";
+  remove_outdated_citations($dba, $var_ad, $pub_ad, $citation_attribs, $citations_pheno_feature, $citations_pheno_feature_attrib);
```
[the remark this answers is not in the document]
nuno-agostinho: Added.

scripts/import/import_EPMC.pl
nuno-agostinho: Updated the code: after removing a citation, it will remove the 'Cited' attribute for a variant if the variant is not in `variation_citation` anymore.

scripts/import/import_EPMC.pl
```diff
@@ -1227,6 +1227,18 @@ sub remove_outdated_citations {
...
+  my $rm_cited_evidence = qq[
+        update %s v
+        set v.evidence_attribs = NULLIF(TRIM(BOTH ',' FROM REPLACE(
+            CONCAT(',', v.evidence_attribs, ','), ',$cited_attrib_id,', ',')), '')
+        where v.variation_id = ?
+          and not exists (select * from variation_citation vc
+                          where vc.variation_id = v.variation_id)
+  ];
+  my $rm_variation_cited_evidence_sth = $dbh->prepare(sprintf $rm_cited_evidence, 'variation');
+  my $rm_variation_feature_cited_evidence_sth = $dbh->prepare(sprintf $rm_cited_evidence, 'variation_feature');
```
dglemos, line 1240: The script is not running:
```
Global symbol "$dbh" requires explicit package name at scripts/import/import_EPMC.pl line 1239. 
Global symbol "$dbh" requires explicit package name at scripts/import/import_EPMC.pl line 1240.
```

nuno-agostinho: Fixed, sorry!

dglemos:
While testing the full dataset, I noticed there was a failure in the method `process_phenotype_feature_attrib()`. 
When running the query on [phenotype_feature_attrib table](https://github.com/Ensembl/ensembl-variation/blob/1e18e968fd5b42ead7908762e68a900fdeb3ecdb/scripts/import/import_EPMC.pl#L1130), some rows are from COSMIC but we don't support citations from COSMIC especially because they are linked to a gene, not a variant.
The query has to be updated to specify the source_id in the query e.g. ClinVar
This wasn't happening before, maybe more data is being imported from COSMIC phenotypes now.

dglemos:
The [query](https://github.com/Ensembl/ensembl-variation/blob/1e18e968fd5b42ead7908762e68a900fdeb3ecdb/scripts/import/import_EPMC.pl#L1130) should be:
```
select pfa.phenotype_feature_id, pfa.value, pf.source_id, pf.object_id
from phenotype_feature_attrib pfa
inner join phenotype_feature pf on pfa.phenotype_feature_id = pf.phenotype_feature_id
join attrib_type att on pfa.attrib_type_id = att.attrib_type_id
left join source s on s.source_id = pf.source_id
where att.code = 'pubmed_id' and s.name = 'ClinVar';
```

dglemos (approved):

PR #917, opened 2022-11-28
Print PolyPhen-2 pipeline errors in eHive

nuno-agostinho:
Since changing ProteinFunction pipeline to run PolyPhen-2 via Singularity, any errors raised only state that the Singularity instance failed. This PR intends to properly display the errors from PolyPhen-2 as before it was run in Singularity.


likhitha-surapaneni (approved):
Looks good to me. Thank you for this enhancement @nuno-agostinho

PR #921, opened 2022-12-06
Update import_vcf script for EVA

dglemos:
- Add a command to update the source version
- Skip synonyms insertion if there are multiple rs ids
- Add evidence attribs to variation and variation feature when a variant has multiple values in `SID`

Details on how to test can be found here: [ENSVAR-5284](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-5284)


jamie-m-a (approved):
Tested with new mouse data, seems to be actng as intended

PR #925, opened 2022-12-08
Improve COSMIC import by chromosome and cleanup

nuno-agostinho:
- Filter COSMIC import by chromosome (no need to split input file beforehand)
- When cleaning up, move tmp tables to new database with same name suffixed by `_tmptables`

Minor improvements:
- Allow to use `cosmic` argument instead of `version` to specify COSMIC version (to avoid confusing with Ensembl release version)
- Fix `cleanup` instructions
- Raise error if cleaning up without providing a registry file
- Print messages with progress of cleanup

### DIFF
--- a/scripts/import/import_cosmic.pl
+++ b/scripts/import/import_cosmic.pl
@@ -91,19 +91,21 @@ sub performCleanup {
     $main_table =~ s/[0-9]+$//;
     next if $table eq $main_table;
 
+    warn "Moving $table contents to $main_table...\n";
     $dbVar->do("INSERT INTO $main_table SELECT * FROM $table;");
     delete_table($dbVar, $table);
   }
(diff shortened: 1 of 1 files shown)


dglemos (commented):
The update looks good!
Is there any particular reason why the option `version` is now called `cosmic`? Using `version` is easier to understand we have to input the version number.

nuno-agostinho:
Hey @dglemos, it was just to make it clear it was related with `cosmic`, not the Ensembl release. But yeah, after using it, it feels weird... I am going to revert that part.

dglemos:
I hadn't thought about the Ensembl release, someone could mix the two that's true. But let's keep it version for now and we'll update it if necessary.

dglemos (approved):

PR #927, opened 2022-12-13
Update citation import script 

dglemos:
When checking for the dbSNP citations:
- truncate title if length > max
- remove <> from title
- fix authors encoding


olaaustine (approved):

olaaustine:
Merged into main and release/110

PR #929, opened 2022-12-14
Update schema document for e109

nakib103:
The changes can be seen in the sandbox - http://wp-np2-11.ebi.ac.uk:7070/
Only schema version have been changed


nuno-agostinho (approved):
Version changed to 109 as expected in schema documents.

PR #930, opened 2022-12-15
Skip deprecated ExAC plugin from doc

nakib103:


nuno-agostinho (approved):
Thanks, @nakib103, seems great! Would be nice in the future if the script skipped plugins not on the config file as well. Added this point to ticket [ENSVAR-5395](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-5395).

nuno-agostinho:
Merged to `main` and `release/109`.

PR #932, opened 2022-12-16
Fix SIFT database version in SIFT/PolyPhen-2 Nextflow pipeline (e110)

nuno-agostinho:
- Use UniRef release note file to get proper database version
- Add default paths to data for SIFT and PolyPhen-2
- Point to `ensemblorg`'s SIFT Docker image


likhitha-surapaneni (approved):
LGTM!

PR #934, opened 2022-12-19
Update vc fixes

jamie-m-a:
Details in here: https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-5397 

Should address the issues encountered with update mode of variation consequence pipeline


diegomscoelho (commented):
Just removed a few objects and gave a small suggestion about deleted file usage.

PR #896, opened 2022-10-03
Add patches for 110

dglemos:
Schema update: sql/patch_109_110_a.sql
Add DDG2P attrib_id to variation_citation: sql/patch_109_110_b.sql ([ENSVAR-4990](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-4990))
Add new clinical significance values: sql/patch_109_110_c.sql ([ENSVAR-5196](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-5196))

PR #935, opened 2022-12-22
Schema and test db patches for e110

diegomscoelho:
## Description

Meta.txt and patches for ensembl repos.

PS. Our patches .sql patches were added previously.
https://github.com/Ensembl/ensembl-variation/pull/896


jamie-m-a (approved):
Looks good to me

PR #945, opened 2023-01-17
feat: remove YPAR skip conditions

diegomscoelho:
Card: [here](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-5407)
## Description

Last dbSNP imports we could not deal with YPAR regions. Being that we only added YPAR variants to X chromosome. Now it should be added to both chromosomes, X and Y.

## Test

Check JIRA card for recent dbSNP import runnings;


dglemos (approved):

dglemos:
Merged into release/110 and main

PR #948, opened 2023-01-23
Fix impact for a consequence

nakib103:
The impact of 'regulatory_region_ablation' should be MODIFIER instead if MODERATE.

http://wp-np2-11.ebi.ac.uk:7070/info/genome/variation/prediction/predicted_data.html


jamie-m-a (approved):
Looks good to me

PR #951, opened 2023-01-31
Fix ranking of feature elongation and truncation consequences

nuno-agostinho:
* Fix ranking of `feature_elongation` and `feature_truncation` consequences
* Update `Config.pm`

Tested locally and working.


diegomscoelho (approved):
LGTM!

PR #956, opened 2023-02-09
fix: More generic rules to add variation_set_id

diegomscoelho:
## Description

When a variant has two values in variation_feature table: 1) One with empty variation_set_id + 2) Filled variation_set_id, in current UnitProt update this will crash. This PR assumes that filled variation_set_id is the valid one and add filled + new variation_set_id in both cases.

## Test

1) Try running update with warning message and comment inserts;

### DIFF
--- a/scripts/import/import_variant_synonyms
+++ b/scripts/import/import_variant_synonyms
@@ -367,12 +367,17 @@ sub add_set_uniprot {
       )
     };
 
-    $db_adaptor->dbc->do($ins_stmt); 
+    # $db_adaptor->dbc->do($ins_stmt); 
     $variation_set_id = $db_adaptor->dbc->db_handle->{'mysql_insertid'};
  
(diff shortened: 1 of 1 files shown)


scripts/import/import_variant_synonyms
```diff
@@ -402,20 +407,11 @@ sub add_synonyms_uniprot {
...
-    WHERE
-      variation_id=?
-  }; 
-
   my $stmt_update = qq{
     UPDATE
       variation_feature 
     SET 
-      variation_set_id = CONCAT_WS(',',variation_set_id,'$variation_set_id')  
+      variation_set_id = '?'
```
dglemos, line 414: It should be `variation_set_id = ?`

dglemos (approved):

dglemos:
Merged into release/110 and main

PR #958, opened 2023-02-15
feat: dbSNP import - Add other + cleaning

diegomscoelho:
## Description

We still do not include refsnp-other in pipeline and do not delete files after import is finished. This PR implement both features to dbSNP-import pipeline.

## Test

1) Create DB + run dbSNP import with:
* chr_other folder inside `split-src` folder.
* using flag `--clean 1` and check if split-src files were deleted when pipeline finishes.


olaaustine (approved):
Tested and the clean option works. 
Thanks @diegomscoelho

olaaustine:
Merged into main and release/110

PR #962, opened 2023-02-23
Update TranscriptVariationAllele - revert PR 814

dglemos:
A user reported an issue in the HGVS protein returned by the plugin ProteinSeqs ([ticket](https://github.com/Ensembl/ensembl-vep/issues/1350))
The HGVSp is not the same as the one returned by VEP.

The issue is caused by this code update: https://github.com/Ensembl/ensembl-variation/pull/814/files 
`$self->{variation_feature_seq}` is updated to store the original alt allele however this has a downstream impact -> the hgvs protein calculated by the plugin is based on the **original alt allele** but the hgvs protein calculated by VEP is calculated based on the **shifted alt allele**. 

The plugins are going to be updated to not use `variation_feature_seq`.

[ENSVAR-5501](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-5501)


diegomscoelho (approved):
LOTM!

PR #938, opened 2023-01-04
Update VariationFeatureAdaptor HGVSc multi-exon

dglemos:
[ENSVAR-5314](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-5314)
Variant Recoder dies with input NM_003629:c.1386_*3del (GRCh37)
The bug is in VariationFeatureAdaptor, it only happens when transcript has no version and it's NM_

PR #963, opened 2023-03-01
Update VariationFeatureAdaptor to support HGVSp multi-exon

dglemos:
When the input is an HGVS protein the method `get_reference()` only considers that the genomic `start` and `end` are within the same exon which is not true for some cases.
More details in: [ENSVAR-2537](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-2537)

This PR is similar to https://github.com/Ensembl/ensembl-variation/pull/938


nuno-agostinho (commented):
Hey @dglemos, the logic seems to work for some examples of HGVSp, thanks! I only have a minor comment.

nuno-agostinho (approved):
Thanks, @dglemos! The PR seems to be working for the variants in [ENSVAR-2537](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-2537).

PR #965, opened 2023-03-03
feat: dbSNP import - Bulk load + sort + dump

diegomscoelho:
## Description

...

## Test

...


olaaustine (approved):
Used in importing for GRCh37. Works great

PR #966, opened 2023-03-03
SIFT/PolyPhen-2 Nextflow pipeline improvements

nuno-agostinho:
- Use increasing amounts of memory when jobs fail because of memory limits
- Retry jobs when LSF job submission fails
- Use tuples to ensure expected pair of arguments (giving two ordered arguments is not safe, as Nextflow may reorder their values when running)
- Increase default number of simultaneous jobs
- Only show jobs as failed if they effectively failed (jobs that finished successfully without output are considered successful now, although downstream jobs will not be run given that there is no output)

### DIFF
--- a/nextflow/ProteinFunction/nextflow.config
+++ b/nextflow/ProteinFunction/nextflow.config
@@ -4,8 +4,9 @@ profiles {
   lsf {
     executor {
       name            = 'lsf'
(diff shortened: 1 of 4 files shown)


likhitha-surapaneni (approved):
Looks good to me, pipeline runs as expected.

PR #972, opened 2023-03-15
Fix long allele_string for variant located after chromosome end

nuno-agostinho:
[ENSVAR-2554](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-2554): Fixes https://github.com/Ensembl/ensembl-rest/issues/397 

## Description

When using VEP with a variant located after chromosome end and a FASTA file, a long allele_string of multiple `NNNNN`s is returned for `JSON` or `VCF` output (but not `TAB` or default `VEP output`).

Similar to how VEP behaves without a FASTA file, this PR intends to return only one `N` in these situations.

## Testing

The VCF (`--vcf`) and JSON (`--json`) output should be the same when using VEP with or without a FASTA file.

Example of running VEP for a variant located after chromosome end:

``` 
perl $vep --id '11:g.148657042delT' --hgvs \
          --cache $cachedir --cache_version 108 --db_version 108 \
          --fasta $fasta \
          --vcf -o vep.vcf
```

### DIFF
--- a/modules/Bio/EnsEMBL/Variation/Utils/FastaSequence.pm
+++ b/modules/Bio/EnsEMBL/Variation/Utils/FastaSequence.pm
@@ -616,6 +616,13 @@ sub _raw_seq {
       $start = 1;
     }
 
+    if ($start > $fa_length) {
+      print STDERR "SLICE HAS START > SEQ LENGTH\n" if $DEBUG;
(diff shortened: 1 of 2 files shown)


nakib103 (approved):
Works as expected!~

PR #973, opened 2023-03-20
Match attrib_type entries with dc value

nakib103:
Some attrib type values in `attrib_entries.sql` have small mismatch with the value in production services. This generally cause DC failure if the attrib tables are generated from this file.

This PR match the values with what we have in production services.


nuno-agostinho:
Hey @nakib103, the changes seem fine. However, is there a way to get the most recent changes directly from Production attributes table?

PR #971, opened 2023-03-14
Update script import_vcf.pl to import all types of chromosomes

dglemos:
1. Update import_vcf.pl to import all chromosomes in the same run even the chr synonyms;
2. Add option `--output_file` to write how many lines were skipped - prints warning with the total of skipped lines. This can be useful for the pipeline;
3. Add option `--sort_vf` to sort the table variation_feature after import;
4. Remove old comments;
5. Remove the progress bar prints.
6. Fix issue when reading a chromosome with semi-colon (example to test: dog)

The script was tested on LSF and SLURM.

PR #980, opened 2023-03-30
Add EVA nextflow pipeline

dglemos:
Create a pipeline to import variants from EVA.
Details: ENSVAR-5502

This PR requires the following updates:
- script import_vcf.pl (https://github.com/Ensembl/ensembl-variation/pull/971)
- variation synonyms (https://github.com/Ensembl/ensembl-variation/pull/977)
- modules/Bio/EnsEMBL/Variation/Utils/Reports.pm (https://github.com/Ensembl/ensembl-variation/pull/982)
- script to copy tables (https://github.com/Ensembl/ensembl-internal-variation/pull/118)


nextflow/EVAImport/main.nf
```diff
@@ -0,0 +1,301 @@
...
+  val registry
+
+  output: val 'ok'
+
+  exec:
+
+  // variable my_set has to be before any if statement
+  // related to this issue: https://github.com/nextflow-io/nextflow/issues/804
+  def my_set = set_names.get(species)
+  for (String name : my_set) {
```
likhitha-surapaneni, line 226: Does it make sense to load the sets in parallel from db load perspective? Can we use [Channels](https://www.nextflow.io/docs/latest/channel.html#fromlist) to do the same?

nextflow/EVAImport/main.nf
```diff
@@ -0,0 +1,301 @@
...
+
+  if(set_names[params.species]) {
+    run_variation_set(run_variant_synonyms.out, var_set_script, files_path, filenames, set_names, params.species, params.registry)
+    run_variation_set_2(run_variation_set.out, var_set_script_2, params.species, params.registry)
+  }
+
+  if(params.citations_file) {
+    run_citations(run_variant_synonyms.out, file(citations_script), params.species, params.registry, params.citations_file)
+  }
+}
```
likhitha-surapaneni, line 301: Can we have a `workflow.onComplete` code block upon completion of the workflow similar to this [example](https://github.com/Ensembl/ensembl-variation/blob/18a12b68c7fd8302744b0e99b105c8d2e6f05e92/nextflow/ProteinFunction/main.nf#LL186C1-L186C20)?

likhitha-surapaneni (changes requested):

nextflow/EVAImport/main.nf
dglemos: The variant sets only start when variation synonyms is finished.

likhitha-surapaneni (approved):
Looks good to me. Request changes have been added. 
Thanks @dglemos

nextflow/EVAImport/main.nf
```diff
@@ -0,0 +1,315 @@
...
+  input:
+  val copy_script
+  val host
+  val dbname
+  val old_dbname
+
+  output: val 'ok'
+
+  script:
+  def new_host = "${host}-ensadmin"
```
nuno-agostinho, line 257: After reviewing https://github.com/Ensembl/ensembl-internal-variation/pull/118, I noticed the `-ensadmin` here. I do not think this is safe because when running this pipeline, we never explicitly give the user/password credentials to allow it to write the database.

I think that it would be better to get user/password parameters as other pipelines do.

When starting the EVA Nextflow pipeline, you can use the shortcut you used here to automatically define the parameters with `${host}-ensadmin details`.

nextflow/EVAImport/main.nf
```diff
@@ -0,0 +1,315 @@
...
+
+  script:
+
+  """
+  perl ${citations_script} -load_file ${file} -registry ${registry} -species ${species}
+  """
+}
+
+
+workflow {
```
nuno-agostinho, line 282: Currently, the pipeline does not allow relative file paths for `params.registry` and `params.input_file`. You can support this by moving the `command_to_run` section of the code here:
```suggestion
workflow {
  // Build command to run EVA import
  registry = file(params.registry)
  input_file = file(params.input_file)

  command_to_run = " -i ${input_file} --source ${params.source} --source_description '${params.description}' --version ${params.version} --registry ${registry} --species ${params.species} --skip_tables '${params.skip_tables}'"

  log.info """
    Importing EVA data with the following parameters: \
    ${command_to_run}
  """
```

In this case, you would also need to use `registry` instead of `params.registry` as the arguments for functions in this workflow.

nuno-agostinho (changes requested):
Hey @dglemos, thanks for the update! While running the pipeline, I had found some issues that could be nice to solve. Anyway, the pipeline seems to run well!

nextflow/EVAImport/nextflow.config
```diff
@@ -0,0 +1,41 @@
+profiles {
+  standard { 
+    process.executor = 'local'
+    process.memory = '2GB'
+    process.cpus = 1
+  }
+  
+  lsf {
```
nuno-agostinho, line 8: You can also add a `slurm` profile similar to `lsf`.

nextflow/EVAImport/main.nf
dglemos: Updated the script.

nextflow/EVAImport/main.nf
```diff
@@ -0,0 +1,315 @@
...
+  prepare_tables(copy_tables_script, params.host, params.dbname, params.old_dbname)
```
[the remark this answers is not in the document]
dglemos: Here we only need the host - the script copy_tables_eva.sh copies tables within the same server. 
It's done this way because in prod-1 we usually have the previous db version.

nextflow/EVAImport/nextflow.config
```diff
@@ -10,6 +10,13 @@ profiles {
     process.memory = '2GB'
     process.cpus = 1
   }
+
+  slurm {
+    process.executor = 'slurm'
+    process.memory = '2GB'
+    process.cpus = 1
+    process.time = 48
```
nuno-agostinho, line 18: Just to make this easier to understand:

```suggestion
    process.time = '48h'
```

nuno-agostinho (changes requested):
Thank you for the changes @dglemos! I can approve/merge it after a small change. Thanks!

nuno-agostinho:
Thanks for the changes, @dglemos, all seems fine! :)

nuno-agostinho (approved):

nuno-agostinho:
Merged to `release/111` and `main`.

PR #986, opened 2023-05-11
Re-format vcf config according to eva integration script

nakib103:
[ENSVAR-5604](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-5604)


nuno-agostinho (approved):
Thank you @nakib103, looks good to me!

Tell me if I can already merge this or if we should wait for some EVA pipeline updates.

PR #988, opened 2023-05-16
Variation docs: sort similar clinical significance terms together

nuno-agostinho:
This PR allows to group similar clinical significance terms together in Variation docs.

See: http://wp-np2-11.ebi.ac.uk:6070/info/genome/variation/phenotype/phenotype_annotation.html


nakib103 (approved):
LOTM!~

PR #991, opened 2023-05-17
Changes to the test db for branching 

olaaustine:
Branching for release/111


nakib103 (changes requested):
Hi, Ola
There should be also change to multi/ontology tables.
see - https://github.com/Ensembl/ensembl-variation/pull/813/files

nakib103 (changes requested):
Hi, Ola
There should be also change to multi/ontology tables.
see - https://github.com/Ensembl/ensembl-variation/pull/813/files

nakib103 (approved):
LGTM!

PR #997, opened 2023-05-31
Create MaveDB plugin data (Nextflow)

nuno-agostinho:
[ENSVAR-5485](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-5485): Nextflow pipeline to create MaveDB plugin data

This pipeline creates a TSV file with variants and associated MaveDB scores. The scores are downloaded from MaveDB and are mapped to variants based on MaveDB mapping files. The mapping files can use two types of HGVS:
- **HGVSg** that contain **genomic** variant coordinates (we need to liftover to hg38 in some cases)
- **HGVSp** that contain **protein** variant coordinates (we need to run Variant Recoder to map them to genomic coordinates)

## Pipeline steps

1. For each MaveDB mapping file, check if corresponding MaveDB scores file is open-access (licence CC0) using MaveDB API
2. Split MaveDB mapping files by HGVS type: either **HGVSg** or **HGVSp**
3. If HGVSg:
	1. Map MaveDB scores to genomic variants using MaveDB mappings file
        - Scores file is automatically downloaded using MaveDB API
	3. LiftOver to hg38 (if needed)
4. If HGVSp:
	1. Get all unique HGVSp from MaveDB mappings file
	2. Run Variant Recoder (VR) to get possible genomic coordinates for HGVSp
	    - Can take up to 6 hours + 70 GB of RAM for a single run with many HGVSp
	    - Given that it uses the online Ensembl database, it may fail due to too many connections
	4. Map MaveDB scores to genomic variants using VR output and MaveDB mappings file
	    - Scores file is automatically downloaded using MaveDB API
5. Concatenate all output files into a single file
6. Sort, bgzip and tabix

You can find [a Mermaid diagram in the README.md file](https://github.com/nuno-agostinho/ensembl-variation/blob/process-mavedb-scores/nextflow/MaveDB/README.md).

Note: the MaveDB API may return `502: Proxy error` when under stress, resulting in failed jobs.

## Variant Recoder improvements

- Support parsing HGVSp with silent changes (such as `NP_003336.1:p.Gly3=` and `NP_003997.1:p.Ter3686=`)
- Parse simple HGVSp deletions-insertions (such as `NP_003336.1:p.159delinsGln`)

## Testing

`./main.nf -profile lsf -resume --mappings [folder_with_MaveDB_mappings]`

You can grab some MaveDB mappings from the URL mentioned in the JIRA ticket description: [ENSVAR-5485](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-5485)


nextflow/MaveDB/nf_modules/licences.nf
```diff
@@ -0,0 +1,74 @@
...
+  res = urllib.request.urlopen(url).read()
+  res = json.loads(res)
+
+  f = open("LICENCE.txt", "w")
+  f.write(res['license']['shortName'])
+  f.close()
+  """
+}
+
+workflow get_files_by_licence {
```
nakib103, line 25: should we have workflow in a module file? should it be best to have 2 modules instead of calling a module from this workflow?

nextflow/MaveDB/nf_modules/licences.nf
```diff
@@ -0,0 +1,74 @@
...
+      }
+    // Return files with matching licences
+    filtered = get_licence.out.
+      filter{ licences.contains(it.last().text) }.
+      map{ it.first() }
+  emit:
+    filtered
+}
+
+def split_by_mapping_type (files) {
```
nakib103, line 45: this function does not seems to be related to license - should we have a separate module for it?  or, if we change the module to name to `fetch` (or something similar - see below comment) move the `get_license` to different module (e.g. - `utils`)?

nextflow/MaveDB/nf_modules/variant_recoder.nf
```diff
@@ -0,0 +1,35 @@
+process get_hgvsp {
```
nakib103, line 1: similar as `split_by_mapping_type()` - we can have separate `utils` modules for them?

nextflow/MaveDB/main.nf
```diff
@@ -0,0 +1,108 @@
...
+    --round     Decimal places to round floats in MaveDB data (default: 4)
+
+  Note: the mappings must be named with the MaveDB accession, such as
+        '00000001-a-1.json'.
+  """
+  exit 1
+}
+
+// Module imports
+include { get_files_by_licence;
```
nakib103, line 44: can we have a more meaningful name instead of `licenses`? how about `fetch` as the main function is to fetch files?

nakib103 (approved):
Thanks nuno for the change, LOTM!~

PR #1004, opened 2023-06-09
Bugfix/vc update mode

jamie-m-a:
Various fixes for the Update mode of variation consequence pipeline - details in https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-5397

### DIFF
--- a/modules/Bio/EnsEMBL/Variation/Pipeline/UpdateVariationFeature.pm
+++ b/modules/Bio/EnsEMBL/Variation/Pipeline/UpdateVariationFeature.pm
@@ -1,7 +1,7 @@
 =head1 LICENSE
 
 Copyright [1999-2015] Wellcome Trust Sanger Institute and the EMBL-European Bioinformatics Institute
-Copyright [2016-2022] EMBL-European Bioinformatics Institute
+Copyright [2016-2023] EMBL-European Bioinformatics Institute
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


dglemos (commented):
There is an issue with variation_feature.consequence_types and new transcripts.
If there is a new transcript in a region with other transcripts already in transcript_variation, the variation_feature consequence is going to only include the consequences from the new transcript. It is ignoring the consequences that were already in transcript_variation.

jamie-m-a:
Good spot @dglemos - you are correct it was missing other overlapping transcripts. I have updated the logic and I believe it's suitable now.

modules/Bio/EnsEMBL/Variation/Pipeline/UpdateVariationFeature.pm
```diff
@@ -59,78 +58,124 @@ sub run {
...
+                            $consequences .= $overlap->[$i][$j] . ',';
+                        }
+                    }
+                    my @csqs = split',',$consequences;
+                    my %get_unique = map{$_ => 1 } @csqs;
+                    my @unique_csqs = keys %get_unique;
+                    my $csq_unique = join(",", @unique_csqs);
+
+                    $dbc->do(qq{
+                            INSERT INTO $temp_table (variation_feature_id, consequence_types)
```
dglemos, line 144: This step is failing in some cases.
[L101](https://github.com/Ensembl/ensembl-variation/blob/d958315ccb1534131c393ba2fee122c098151564/modules/Bio/EnsEMBL/Variation/Pipeline/UpdateVariationFeature.pm#L101) inserts the new consequence types into the temp table. 
If a gene has updated/new transcripts + deleted transcript this insertion (L144) is going to try to insert the same row into the temp table but it fails due to duplicated keys.
Probably an INSERT IGNORE INTO would be enough to fix the error?

dglemos (approved):

dglemos:

PR #1005, opened 2023-06-12
Update HGMD script

dglemos:
Replace `$slice->get_seq_region_id;` this method does not work correctly.
As HGMD has the same chr names as Ensembl we can fetch the seq regions directly.


likhitha-surapaneni (approved):
Looks good to me. Thank you @dglemos!

PR #1012, opened 2023-07-03
Increased default analysis_capacity for CAR pipeline

likhitha-surapaneni:
DumpHGVS default analysis_capacity increased to 500
GetCAR default analysis_capacity increased to 30


nakib103 (approved):
LOTM!~

PR #1016, opened 2023-07-20
Fixed ref codon and ref peptide used for HGVSp

likhitha-surapaneni:
Ticket: [ENSVAR-778](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-778)

### Issue
 Codon and cds for reference allele are incorrect resulting in incorrect peptide and HGVSp

### Testing

```
3:g.41266031_41266626del
```


likhitha-surapaneni:
Closing due to duplicated PR

PR #994, opened 2023-05-24
Some (more) HGVS updates

nakib103:
- [ENSVAR-5765](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-5765): VEP incorrectly term HGVS type as ins when it should be dup. This happens when the variant strand is in reverse and transcript strand is forward. 
- [ENSVAR-5625](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-5625): We are clipping alleles when the HGVS type is `delins` and update type and co-ordinates. We are not considering type changing to `dup` currrently. Related issue - https://github.com/Ensembl/ensembl-vep/issues/500 
- [ENSVAR-2140](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-2140): Add new argument `--hgvsp_use_prediction` to output HGVSp in prediction format. Related issue - https://github.com/Ensembl/ensembl-vep/issues/861

_Test data are in the respective JIRA tickets_

PR #1017, opened 2023-07-21
Fix warning when there is no pre-seq

nakib103:
When there is no "prefixed sequence" between ref and alt alleles that matches a warning is produced by this line -

https://github.com/Ensembl/ensembl-variation/blob/2fc4a6a5b931f4dc80266b4867cef8fb8a4b6d2e/modules/Bio/EnsEMBL/Variation/TranscriptVariationAllele.pm#L2187

The issue was introduced in this PR - https://github.com/Ensembl/ensembl-variation/pull/994
and result of using undefined `$preseq` - 

https://github.com/Ensembl/ensembl-variation/blob/2fc4a6a5b931f4dc80266b4867cef8fb8a4b6d2e/modules/Bio/EnsEMBL/Variation/TranscriptVariationAllele.pm#L2118


dglemos (approved):

dglemos:

PR #1022, opened 2023-08-01
Fix SV import script

nakib103:
- there were undefined variable in use - `$dba`; it should variation database adaptor - `$vda`
- In `post_processing_sample` we only have logic for `DGVa` study. While running the script other study generates warnings in this function. It is best to put a condition to run this function only when we are importing `DGVa`.

### DIFF
--- a/scripts/import/import_all_sv_data.pl
+++ b/scripts/import/import_all_sv_data.pl
@@ -274,7 +274,7 @@ foreach my $in_file (@files) {
 # Post processing for mouse annotation (delete duplicated entries in structural_variation_sample)
 post_processing_annotation() if ($species =~ /mouse|mus/i);
 post_processing_feature();
-post_processing_sample();
+post_processing_sample() if ($source_name eq 'DGVa');
 post_processing_phenotype();
 post_processing_clinical_significance();
 post_processing_failed_variants();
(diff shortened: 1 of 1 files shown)


scripts/import/import_all_sv_data.pl
```diff
@@ -274,7 +274,7 @@ =head1 CONTACT
 # Post processing for mouse annotation (delete duplicated entries in structural_variation_sample)
 post_processing_annotation() if ($species =~ /mouse|mus/i);
 post_processing_feature();
-post_processing_sample();
+post_processing_sample() if ($source_name eq 'DGVa');
```
dglemos, line 277: We also want to run it for mouse.

nakib103: Yeah, then I realized the 2 queries we are running are both for `DGVa` studies -

https://github.com/Ensembl/ensembl-variation/blob/d34d25e481f22c11705b23a1aded34feb8434010/scripts/import/import_all_sv_data.pl#L1777
https://github.com/Ensembl/ensembl-variation/blob/d34d25e481f22c11705b23a1aded34feb8434010/scripts/import/import_all_sv_data.pl#L1794

So, I think that condition is redundant.

dglemos (approved):

scripts/import/import_all_sv_data.pl
nakib103: Keeping the condition logic as it is for now in line with offline discussion

PR #1023, opened 2023-08-01
VEP handling Y PAR variants

dglemos:
[ENSVAR-5478](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-5478)

When we input an hgvs with the gene name, VEP returns output for chrY but it should be chrX.

Input
`PLCXD1:c.57_61del`


jamie-m-a (approved):
Tested, works to fix issue on 110, no longer needed for 111 but doesn't impact normal function.

PR #1024, opened 2023-08-03
Update citations import - remove die

dglemos:
In release 111, the import died: `"No variation found for '$variation_id'!\n"`
There were new dbSNP + citation tables that were copied from the previous release.
Some variants were not included in 111 anymore so the script was killed when the variant was not found. For this specific case, it was expected some variants not be there anymore.
This PR replaces the die statement with a print.
We should let the script finish and investigate why/if some variants/publications were not found.


nuno-agostinho (approved):
Seems to work as expected, thanks @dglemos!

nuno-agostinho:
Merged to `main` and `release/111`.

PR #1025, opened 2023-08-03
Import SV script fix - header label corruption

nakib103:
The header parser function `get_header_info` was doing greedy regex match until `:` found for getting label. It results in corrupted label if the header line includes a second (or third ...) `:`. For example in `nstd207` study the $study_desc is coming from - 

```
# Study_description: Long-read and strand-specific sequencing technologies together facilitate the de novo assembly of high-quality haplotype-resolved human genomes without parent-child trio data. We present 64 assembled haplotypes from 32 diverse human genomes. These highly contiguous haplotype assemblies (average contig N50: 26 Mbp) ...
```

We should do a lazy match instead.


dglemos (approved):

dglemos:

PR #1026, opened 2023-08-15
Detect input format using VEP code

nuno-agostinho:
1. Update regex for input data detection based on VEP.
2. Create new exported function for input format detection to be used by VEP, avoiding code redundancy.
3. Accept CAID and Region input in web VEP.

## Ignore alternative allele when checking VCF format

The idea of this function is to check if this format is VCF: only these columns are needed to check that.

Basically, we should not check for the alternative allele here because:

- The valid characters for the alternative allele are constantly changing with new versions of the [VCF specification](https://samtools.github.io/hts-specs/VCFv4.4.pdf)
- It only checks the first line (other lines may have issues and are ignored)
- The regex may not be up-to-date (it already happened that we support a specific type of alternative allele but forgot to update the regex here)
- If the alternative allele is not supported, it should be the function [`create_VariationFeatures()` from `Parse::VCF`](https://github.com/Ensembl/ensembl-vep/blob/611114ef13d7bcbcfd330894711749c9a2030b04/modules/Bio/EnsEMBL/VEP/Parser/VCF.pm#L242) to return that issue with a warning (and not an error like `detect_format()`, because the format is still VCF).
- We get user-reported issues related with the strict regex here, even though the line is fully supported by VEP

## Related PRs

- https://github.com/Ensembl/ensembl-vep/pull/1477
- https://github.com/Ensembl/public-plugins/pull/717

### DIFF
--- a/modules/Bio/EnsEMBL/Variation/Utils/VEP.pm
+++ b/modules/Bio/EnsEMBL/Variation/Utils/VEP.pm
@@ -92,8 +92,6 @@ use Bio::EnsEMBL::Variation::DBSQL::StructuralVariationFeatureAdaptor;
 use Bio::EnsEMBL::Variation::TranscriptStructuralVariation;
 use Bio::EnsEMBL::Variation::Source;
 
-use Bio::EnsEMBL::VEP::Parser qw(detect_line_format);
-
 # we need to manually include all these modules for caching to work
 use Bio::EnsEMBL::CoordSystem;
 use Bio::EnsEMBL::Transcript;
@@ -115,6 +113,8 @@ use vars qw(@ISA @EXPORT_OK);
 @ISA = qw(Exporter);
 
 @EXPORT_OK = qw(
+    &_valid_region_regex
+    &check_format
     &detect_format
     &parse_line
     &vf_to_consequences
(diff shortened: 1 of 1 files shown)


modules/Bio/EnsEMBL/Variation/Utils/VEP.pm
```diff
@@ -382,57 +384,76 @@ sub parse_line {
...
     # VCF: 20  14370  rs6054257  G  A  29  0  NS=58;DP=258;AF=0.786;DB;H2  GT:GQ:DP:HQ
     elsif (
-        $data[0] =~ /(chr)?\w+/ &&
-        $data[1] =~ /^\d+$/ &&
-        $data[3] =~ /^[ACGTN\-\.]+$/i &&
-        $data[4] && $data[4] =~ /^([\.ACGTN\-\*]+\,?)+$|^(\<[\w]+\>)$/i
+        $line[0] =~ /(chr)?\w+/ &&
+        $line[1] =~ /^\d+$/ &&
+        $line[3] && $line[3] =~ /^[ACGTN\-\.]+$/i &&
+        $line[4]
```
nakib103, line 432: Why we are not checking regex for alternate allele?

nakib103 (changes requested):

modules/Bio/EnsEMBL/Variation/Utils/VEP.pm
nuno-agostinho: Hey @nakib103! The idea of this function is to check if this format is VCF: only these columns are needed to check that. I am simply copying here the code from ensembl-vep, already accepted in PR https://github.com/Ensembl/ensembl-vep/pull/1370.

Basically, we should not check for the alternative allele here because:
- The valid characters for the alternative allele are constantly changing with new versions of the [VCF specification](https://samtools.github.io/hts-specs/VCFv4.4.pdf)
- It only checks the first line (other lines may have issues and are ignored)
- The regex may not be up-to-date (it already happened that we support a specific type of alternative allele but forgot to update the regex here)

If the alternative allele is not supported, it should be the function [`create_VariationFeatures()` from `Parse::VCF`](https://github.com/Ensembl/ensembl-vep/blob/611114ef13d7bcbcfd330894711749c9a2030b04/modules/Bio/EnsEMBL/VEP/Parser/VCF.pm#L242) to return that issue with a warning (and not an error like `detect_format()` did, because the format is still VCF).

nakib103: Thanks Nuno,

That makes sense if we do not want to add checks to alt allele if its standard is changing frequently. But in that PR you still had checks for alts in the `detect_format()` -
https://github.com/nuno-agostinho/ensembl-vep/blob/7ecd5d059bb5cd396624804b8fdc0eb8a2f35334/modules/Bio/EnsEMBL/VEP/Parser.pm#L442

Is the change introduced here (not checking alts) intended?

nakib103: Thanks Nuno,

That makes sense if we do not want to add checks to alt allele if its standard is changing frequently. But in that PR you still had checks for alts in the detect_format() -
https://github.com/nuno-agostinho/ensembl-vep/blob/7ecd5d059bb5cd396624804b8fdc0eb8a2f35334/modules/Bio/EnsEMBL/VEP/Parser.pm#L442

Is the change introduced here (not checking alts) intended change for this PR?

nuno-agostinho: Yes, you are right, the PR still checks the alternative allele. Still, I think we should remove because of the reasons mentioned. I added this information (and the reasons why) in the PR description. Thanks!

nakib103 (approved):
Thanks Nuno, looks alright to merge!

PR #1030, opened 2023-08-17
Add start_retained for non-ATG transcript

nakib103:
JIRA Ticket - [ENSVAR-3154](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-3154)
Also see - https://www.ebi.ac.uk/seqdb/confluence/display/EV/VEP+-+non-standard+translation

### Problem
There are some transcript that have non-ATG start, for example - gene WT1, transcript ENST00000332351. It has already been fixed to report `synonymous_variant` if the start codon is altered to an ATG (which is the normal start codon). But we should report `start_retained_variant` instead of `synonymous_variant` as it is more accurate/appropriate.

### Solution
Added another function that checks the amino acid that replaces the start codon and see if is an 'ATG'.

### Test
Example command -
```
vep -i input_data --force --database

#input_data
11	32435360	rsfake	G	T
```

### DIFF
--- a/modules/Bio/EnsEMBL/Variation/Utils/VariationEffect.pm
+++ b/modules/Bio/EnsEMBL/Variation/Utils/VariationEffect.pm
@@ -934,7 +934,12 @@ sub start_retained_variant {
 
     my $pre = $bvfoa->_pre_consequence_predicates;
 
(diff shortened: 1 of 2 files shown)


olaaustine (approved):
Tested with the example and it works and with other variants and it does not change the annotation of non-synonymous variants, tested with other examples.

PR #1035, opened 2023-09-01
Fix MaveDB plugin data creation

nuno-agostinho:
ENSVAR-5960: Ensure MaveDB VEP plugin works with regulatory data

- Fix start position for all genomic HGVS variants (zero-index position is now one-index position to conform with protein HGVS)
- Update URL for MaveDB REST endpoint (previous one is now invalid)

Note: The path to the updated plugin data is mentioned in ENSVAR-5960 and can be tested with https://github.com/Ensembl/VEP_plugins/pull/628.


nakib103 (approved):
LOTM!~

PR #1037, opened 2023-09-07
112 schema patch

jamie-m-a:


nuno-agostinho (approved):
Tested commands in a demo database, everything seems to be working as expected, thanks @jamie-m-a!

PR #1040, opened 2023-09-11
Report consequences for each breakend

nuno-agostinho:
ENSVAR-5962: Report consequences for each breakend

- For variants whose class_SO_term contains `breakpoint`, their breakends are parsed based on the allele string and the result is cached in `$vf->{breakends}`
- For each breakend/feature, one alternative allele (`Bio::EnsEMBL::Variation::StructuralVariationOverlapAllele `) is created if the feature is near the breakend
    - Reference breakend is represented with `N.` (if there is any `left`-positioned breakend) or `.N`
    - Proximity is tested using new function `_close_to_feature()`
    - Improve `Bio::EnsEMBL::Variation::StructuralVariationOverlapAllele` to have a `new()` constructor that includes `symbolic_allele` string (such as `N[8:5455654[`) and a `breakend` with the data from a single breakend from `$vf->{breakends}`
- Simplify `feature_truncation()`

### Testing

Test changes in https://github.com/Ensembl/ensembl-vep/pull/1496


likhitha-surapaneni:
Merged to main and release/112

PR #1042, opened 2023-09-27
Fix lookup direction for dup when sequence is reversed

nakib103:
Related issue - https://github.com/Ensembl/ensembl-vep/issues/1485
JIRA ticket - https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-6048

### The original issue was reported in [JIRA Ticket](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-5765) and (apparently) fixed in this [PR](https://github.com/Ensembl/ensembl-variation/pull/812)

**The Issue**
- variant reported in HGVSc ID with transcript in reverse strand was getting `ins` instead of `dup`

The fix was to -
1. update variation feature start and end co-ordinate. The understanding was that variant will always have strand +1 and so the whatever the input is the we need to change the position to `$start = $end + 1`.

**This is incorrect**: variant can have strand -1. It can happen with HGVS, VEP default, or REST like input - as HGVSc can have transcript in reverse strand and the other two have the option to give strand value (e.g - `2 58386929 58386928 -/AATTA -`)

This PR corrects that mistake. Note that original version also had problem. Assuming `$end < $start` for HGVSc with transcript in reverse strand. So, we change - 
```
$start = $end ;
     if($strand  == 1){ $start++; }
     else{             $end--;   }
```

to,
```
if($strand  == 1){ $end = $start - 1; }
    else{ $start = $end + 1 }
```
This updates the position to the start of the insertion, as this is the same thing that we understood for an insertion described in VEP default / REST like input, e.g - 
`2 58386929 58386928 -/AATTA +` (insertion happening after `58386928` and not `58386933`)

2. For `--shift_hgvs 0` and transcript in reverse strand we updated lookup direction to look at 5' direction (forward) instead of looking at 3' direction (backward) by default.

**This is obsolete**: With the change in this PR, we should always look at 5' direction. So we have removed the logic to pass `lookup_direction` to `hgvs_variant_notation` function.

### Another related bug addressed in [JIRA Ticket](https://github.com/Ensembl/ensembl-variation/pull/994) fixed in [PR](https://github.com/Ensembl/ensembl-variation/pull/994)

**The Issue**
- When variation and transcript strand is opposite we again get `ins` instead of `dup`, e.g - `ENST00000402135.3:c.1111_1114dup`

The fix was to -
1. Should only complement the transcript sequence instead of reverse complement when the variant is reverse strand and transcript is forward strand. 

**This does not solve the root cause**: the fix was only symptomatic and only fixed the test dataset and introduced error in other cases. The original issue is the shifting itself. We shift [two sequence](https://github.com/Ensembl/ensembl-variation/blob/d34d25e481f22c11705b23a1aded34feb8434010/modules/Bio/EnsEMBL/Variation/TranscriptVariationAllele.pm#L289) - 
- `seq-to_check`: that is the sequence in the reference sequence in forward strand. The direction of shift depend on the strand of the transcript
- `hgvs_output_string`: this is the allele derived from the original input. The sequence can be in forward or reverse strand depending on the input. 

The thing is this two sequence can be on different strand, for example, when we have the variation and transcript on different strand as in this issue and you need to account for that when you are shifting. If the strand is opposite the shift direction should be opposite too.


modules/t/hgvs_parser.t
```diff
@@ -544,9 +563,9 @@ my %test_output_no_shift = (
            "ENSP00000435307.1:p.Arg151_Ser152insTer",
             "insertion, stop gained [-1]"
            ],
-     3 =>  ["NC_000013.10:g.51519667_51519668insG",    ##rs17857128
+     3 =>  ["NC_000013.10:g.51519668dup",    ##rs17857128
```
nuno-agostinho, line 566: Remove rsID, this is not the same variant:
```suggestion
     3 =>  ["NC_000013.10:g.51519668dup",
```

modules/Bio/EnsEMBL/Variation/Utils/Sequence.pm
```diff
@@ -569,22 +569,37 @@ sub hgvs_variant_notation {
   # If this is an insertion, we should check if the preceeding reference nucleotides match the insertion. In that case it should be annotated as a multiplication.
   if (!$ref_length) {
```
nuno-agostinho, line 570: Given that we always look at the same code twice (one for each direction), I would suggest to replace the code inside this `if` statement to make it simpler and avoid code redundancy:

```perl
  if (!$ref_length) {
  
    # lookup_direction = 1: look at 5' then 3' direction for ref string (reversed order for -1)
    my @lookup_direction = $lookup_order == 1 ? (1, -1) : (-1, 1);

    for my $direction (@lookup_direction) {
      # Get the same number of nucleotides preceding the insertion as the length of the insertion
      my $prev_str = $direction == 1 ? 
        substr($ref_sequence,$ref_end,$alt_length) : 
        substr($ref_sequence,($ref_end-$alt_length),$alt_length);

      # If they match, this is a duplication
      if ($prev_str eq $alt_allele) {
        $notation{'end'}   = $display_start + $alt_length - 1 if $direction == 1;
        $notation{'start'} = ($display_end - $alt_length + 1) if $direction == -1;
        $notation{'type'}  = 'dup';

        return \%notation;
      } 
    }

    # If they didn't match it's a plain insertion
    $notation{'start'} = $display_end;
    $notation{'end'}   = $display_start;
    $notation{'type'}  = 'ins';

    return \%notation;
  }
```

nuno-agostinho (changes requested):
Thanks for the fix, @nakib103. The code seems to be working as intended, I left some comments, but I can approve the PR if you don't agree with them. Thanks again!

modules/t/hgvs_parser.t
nakib103: updated 👍

modules/Bio/EnsEMBL/Variation/Utils/Sequence.pm
nakib103: Thanks @nuno-agostinho , makes sense, updated!

nakib103:
Thanks @nuno-agostinho , I have updated with change requested.
We should not marge this PR though. As per with the discussion with Jamie, there is another pending [bugfix](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-5920) required for HGVS and these two fix should go together in case there is common code affecting one another.

serge2016:
Hello! What is the ETA for this PR to be merged? Hope to see this very soon...

likhitha-surapaneni (approved):
Thanks @nakib103 , LGTM

nuno-agostinho (approved):

nakib103:
merged to `main` and `postrealeasefix/112`

PR #1043, opened 2023-10-09
Dump GO: add dir to output file path

nakib103:
According to this PR GO plugin only returns the filename using `_prepare_filename()` function, so we need to add the dir to actually have the full output path.


olaaustine (approved):
LGTM! Thank you @nakib103

olaaustine:
Merged into release/112 and main

PR #1044, opened 2023-10-11
VEP docs: Fix examples in sources documentation

nuno-agostinho:
ENSVAR-6045

* Fix examples for VCF-only species: add region information instead of using rsIDs (currently rsIDs are not supported for VCF-only species)
* [Fix examples of structural variant phenotypes](https://github.com/Ensembl/ensembl-variation/commit/1df19c1dd680aea7e7184a89037f1612e8839276)
* [Ignore case of species Latin name](https://github.com/Ensembl/ensembl-variation/pull/1044/commits/bb2a5a3b48097882b22cfda6aa9b00d35c863bab)

Backport to e111?

## Testing

Check changes in my [sandbox](http://wp-np2-11.ebi.ac.uk:6070/info/genome/variation/species/sources_documentation.html).


scripts/docs/sources2html.pl
```diff
@@ -1262,33 +1256,46 @@ sub get_example {
...
+      }
     }
+  } else {
+    # Features from VCF
+    my $file_full_path = $data;
+    my $source_name    = $param;
+
+    # create a variation feature string
+    $example = `bcftools query -f "%CHROM:%POS:%REF-%ALT\n" $file_full_path | head -n 1`;
+    $example =~ s/-/_/;
```
nakib103, line 1291: Should we do this on one query?

```suggestion
    $example = `bcftools query -f "%CHROM:%POS:%REF\_%ALT\n" $file_full_path | head -n 1`;
```

nuno-agostinho: Hi @nakib103, I tried, but it didn't work for me, it tries to fetch `%REF_` instead of `%REF`... Does it work for you?

nakib103: yeah, I needed to escape `_` with a `\_`. I am running the script the same as you had in this PR but when I tested with `\_` it worked fine.

nakib103 (changes requested):

nakib103 (approved):
Looks good now

PR #1047, opened 2023-10-24
Update billion count icon in docs and other updates

nakib103:
- add different color for billion counts in pages that list variant counts
- add chimpanzee and gibbon json in `vcf_config.json` which somehow got missed.
- add `vcf_species_list_111.json` data file
- fix syntax error in clin_significance generation script which causing failure in generating the tables.
- fix missing frequency for human in population page (because we were only checking genotype and ignoring frequency)

### DIFF
--- a/modules/Bio/EnsEMBL/Variation/DBSQL/vcf_config.json
+++ b/modules/Bio/EnsEMBL/Variation/DBSQL/vcf_config.json
@@ -938,6 +938,30 @@
         "source_version": 4
     },
     {
+      "id": "eva_nomascus_leucogenys_gca000146795v3",
+      "description": "Variants from EVA",
+      "species": "nomascus_leucogenys",
+      "assembly": "Nleu_3.0",
+      "type": "remote",
+      "filename_template": "http://ftp.ebi.ac.uk/pub/databases/eva/rs_releases/release_4/by_species/nomascus_leucogenys/GCA_000146795.3/GCA_000146795.3_current_ids.vcf.gz",
+      "use_as_source": 1,
+      "use_seq_region_synonyms": 0,
+      "source_name": "EVA",
+      "source_version": 4
+    },
+    {
+      "id": "eva_pan_troglodytes_gca000001515v5",
+      "description": "Variants from EVA",
+      "species": "Pan_troglodytes",
+      "assembly": "Pan_tro_3.0",
+      "type": "remote",
+      "filename_template": "https://ftp.ebi.ac.uk/pub/databases/eva/rs_releases/release_4/by_species/pan_troglodytes/GCA_000001515.5/GCA_000001515.5_current_ids.vcf.gz",
+      "use_as_source": 1,
+      "use_seq_region_synonyms": 1,
+      "source_name": "EVA",
+      "source_version": 4
+    },
+    {
         "id": "eva_ficedula_albicollis_gca000247815v2",
         "description": "Variants from EVA",
         "species": "Ficedula_albicollis",
--- a/scripts/docs/generate_population_table.pl
+++ b/scripts/docs/generate_population_table.pl
@@ -379,6 +379,10 @@ sub get_vcf_content_types {
     if ( ($info_field =~ /AF=/) || ($info_field =~ /AC=/ && $info_field =~ /AN=/) ) {
       push @types, "frequency";
     }
+    # a hard-coded check for NCBI-ALPHA and TOPMED as they have very special field for frequency
+    if ( ($info_field =~ /AN_SAMN/) || ($info_field =~ /TOPMED=/) ) {
+      push @types, "frequency";
+    }
   }
   
   return @types;
--- a/scripts/docs/generate_population_table.pl
+++ b/scripts/docs/generate_population_table.pl
@@ -181,7 +181,6 @@ my $html_pop = '';
 
 # Loop over the species (placing human first)
 foreach my $species (sort { ($a !~ /Homo/ cmp $b !~ /Homo/) || $a cmp $b } keys(%pops_list)) {
-
   next unless  %{ $pops_list{$species} };
 
   my $id_species = $species;
@@ -374,8 +373,12 @@ sub get_vcf_content_types {
     my $line = `tabix -D $file_full_path $chr | head -n 1`;
   
     my $format_field = (split /\t/, $line)[8];
-    
     push @types, "genotype" if $format_field;
+
+    my $info_field = (split /\t/, $line)[7];
+    if ( ($info_field =~ /AF=/) || ($info_field =~ /AC=/ && $info_field =~ /AN=/) ) {
+      push @types, "frequency";
+    }
   }
   
   return @types;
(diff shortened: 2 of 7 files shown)


scripts/docs/species_list.pl
```diff
@@ -360,14 +361,22 @@ sub round_count {
...
+  # From 100 million to 999.9 million
+  elsif ($count =~ /^(\d{3})\d{6}$/) {
     my $number = $1;
     $count = "$number M";
     $count_label = "Over $number million $type";
     $count_display = $count;
     $bg_class = $colours{'hundred_million'}{'colour'};
   }
+  # From 1 to 9.9 billion
+  elsif ($count =~ /^(\d)(\d)\d{8}$/) {
```
nuno-agostinho, line 373: Wouldn't it be better to future-proof this to also capture here 10 billion and above? Otherwise, if we forget to update this, those big numbers will be printed in full based on the `else` statement (which is worse).

```suggestion
  elsif ($count =~ /^(\d+)(\d)\d{8}$/) {
```

scripts/docs/sources2html.pl
```diff
@@ -1232,14 +1228,22 @@ sub get_count {
...
+  # From 100 million to 999.9 million
+  elsif ($count =~ /^(\d{3})\d{6}$/) {
     my $number = $1;
     $count = "$number M";
     $count_label = "Over $number million $end_label";
     $count_display = $count;
     $bg_class = $colour_class{'hundred_million'};
   }
+  # From 1 billion to 9.9 billion
+  elsif ($count =~ /^(\d)(\d)\d{8}$/) {
```
nuno-agostinho, line 1240: Could we reduce the redundancy here? It's the same function in two different scripts.

scripts/docs/sources2html.pl
```diff
@@ -1318,17 +1322,20 @@ sub get_vcf_content_types {
   my $genotypes = `tabix $file_full_path -H | grep '##FORMAT' | grep 'ID=GT'`;
   push @types, "genotype" if $genotypes;
   
+  my $chr = `tabix $file_full_path -l | head -n 1`;
+  chop $chr;
```
nuno-agostinho, line 1326: Shouldn't we use `chomp` instead of `chop`? `chomp` only removes a newline character whereas `chop` removes the last character, no matter which one it is.

```suggestion
  chomp $chr;
```

scripts/docs/sources2html.pl
```diff
@@ -495,30 +497,17 @@ sub source_table {
...
-          # Set description
           $description = "Variants imported from EVA";
         }
 
-        # Assuming only one config will have use_as_source set per species
-        if ( grep /^genotype$/, @types){
-          # Update source name and url to study id if possible
-          if ($source_name =~ /^(?!PRJ)/){
-            # Try getting the study id from database if not in vcf collection
-            my $sth3 = get_connection_and_query($db_name, $hostname, $sql_display_group);
```
nuno-agostinho, line 507: The study ID is always in the VCF collection now?

modules/Bio/EnsEMBL/Variation/DBSQL/vcf_config.json
```diff
@@ -937,6 +937,30 @@
         "source_name": "EVA",
         "source_version": 4
     },
+    {
+      "id": "eva_nomascus_leucogenys_gca000146795v3",
+      "description": "Variants from EVA",
+      "species": "nomascus_leucogenys",
```
nuno-agostinho, line 943: The species name needs to be capitalised or not really?

```suggestion
      "species": "Nomascus_leucogenys",
```

nuno-agostinho (changes requested):
Hi @nakib103, thanks for the PR! I left some comments.

modules/Bio/EnsEMBL/Variation/DBSQL/vcf_config.json
nakib103: The one place I know it use do `lc` - https://github.com/Ensembl/ensembl-variation/blob/d34d25e481f22c11705b23a1aded34feb8434010/modules/Bio/EnsEMBL/Variation/DBSQL/VCFCollectionAdaptor.pm#L144

Do you know why it would need capitalizing?

scripts/docs/sources2html.pl
nakib103: No, but it is there we want to have the study url instead of EVA url (in which case the source name will be `"EVA"`)

scripts/docs/species_list.pl
nakib103: Quite true, I have updated to include more than 10 billion in this regex.

modules/Bio/EnsEMBL/Variation/DBSQL/vcf_config.json
nuno-agostinho: I am asking because in other species it's capitalised, but as long as it works, I'm fine with it.

scripts/docs/sources2html.pl
nuno-agostinho: Okay, thanks!

scripts/docs/sources2html.pl
nuno-agostinho: Yes, not important for this PR. Thanks!

nuno-agostinho:
Hey @nakib103, all seems fine with this PR, thanks for the changes!

nuno-agostinho (approved):

nuno-agostinho (approved):
Looks good to me, thanks @nakib103!

scripts/docs/generate_population_table.pl
```diff
@@ -379,6 +379,10 @@ sub get_vcf_content_types {
     if ( ($info_field =~ /AF=/) || ($info_field =~ /AC=/ && $info_field =~ /AN=/) ) {
       push @types, "frequency";
     }
+    # a hard-coded check for NCBI-ALPHA and TOPMED as they have very special field for frequency
+    if ( ($info_field =~ /AN_SAMN/) || ($info_field =~ /TOPMED=/) ) {
+      push @types, "frequency";
+    }
```
nuno-agostinho, line 385: Do you think it would be better to put everything in the same `if` statement? Also, should `AN_SAMN` be `AN_SAMN=` instead?

```suggestion
    if (
      ($info_field =~ /AF=/) ||
      ($info_field =~ /AC=/ && $info_field =~ /AN=/) ||
      # hard-coded check for NCBI-ALPHA and TOPMED as they have very special field for frequency
      ($info_field =~ /AN_SAMN=/) || ($info_field =~ /TOPMED=/)
    ) {
      push @types, "frequency";
    }
```

nakib103: inline comment would not work actually.
the actual field have full sample name `AN_SAMN...=` so I did not put the equal sign here.

nuno-agostinho (approved):

nakib103:
merged to `postreleasefix/112`, `release/112`, and `main`

PR #491, opened 2019-11-20
Task99/trim alleles for vcf dumps

at7:
Major updates
-- trim allele_string in gvf2vcf.pl script
-- use FASTA file and ancestral allele file for human only when running gvf2vcf.pl
-- don't compress human GVF dumps before running gvf2vcf.pl. Compress everything after parsing to VCF

Minor updates
-- add debug option
-- cache source information in dump_gvf.pl
-- use methods from BaseVariationProcess.pm instead of BaseDataDumpsProcess.pm

PR #1049, opened 2023-10-24
Update release dump data dir from noah

nakib103:
We have an update to use FASTA and ancestral allele in the release dump pipeline from this commit -
https://github.com/Ensembl/ensembl-variation/pull/491

- The FASTA is used to query sequence instead of using DB
- The ancestral allele file is used to correct AA for indels after changing them in VCF format.
They are not mandatory parameter as only supposed to be used with human. But AA seems to be necessary as we do need correcting of AA for indels.

The location of these files can be passed to the pipeline using `data_dir` parameter. Current default value is of a noah dir - this PR updates it to the equivalent codon dir.

To be able to use this we should update the data_dir with updated files. Check updated docs here -
https://www.ebi.ac.uk/seqdb/confluence/display/EV/Dump+GVF+and+VCF+for+a+new+e%21+and+EG+releases#DumpGVFandVCFforanewe!andEGreleases-Beforerunningthepipeline


jamie-m-a (approved):
Looks good thanks Nakib!

jamie-m-a:
Can we rebase to postreleasefix/112 please @nakib103 - we won't run this fix for 111

marcoooo (approved):

PR #1052, opened 2023-11-07
Update dbSNP script - fix MAF

dglemos:
Update the script `load_dbsnp.pl`
- Fix MAF for SNVs
- Skip indels: JSON has problems with the allele representation; it is better to skip indels for now otherwise the script is going to insert incorrect values into the db

More details: [ENSVAR-6086](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-6086)

To test the code need to run the dbSNP pipeline.


dglemos:
The MAF is not supposed to be imported by dbSNP anymore.
Closing this PR.

PR #1055, opened 2023-11-13
Allow developer other than Ensembl as plugin author

nakib103:
UTRannotator have Ensembl copyright but also external developer as author. (See - https://github.com/Ensembl/VEP_plugins/pull/657)

Current plugin doc creation script does not allow that. This PR removes that restriction.

### DIFF
--- a/scripts/docs/update_web_vep_plugins_documentation.pl
+++ b/scripts/docs/update_web_vep_plugins_documentation.pl
@@ -253,7 +253,7 @@ sub read_plugin_file {
     }
     
     # Contact/developper information
-    if ($line =~ /=head1 CONTACT/ && scalar(@developer) == 0) {
+    if ($line =~ /=head1 CONTACT/) {
       my $contact_flag = 1;
       while ($contact_flag != 0) {
         $line = <F>;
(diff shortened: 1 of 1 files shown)


scripts/docs/update_web_vep_plugins_documentation.pl
```diff
@@ -263,7 +263,10 @@ sub read_plugin_file {
         }
         else {
           $line =~ s/^\s+//;
-          push(@developer, $1) if ($line =~ /(.+)\s+</);
+          if ($line =~ /(.+)\s+</) {
+            # some plugin had old-style Ensembl contact info with lots of text
+            push @developer, $1 unless (grep /^$1$/, @developer || $1 eq "developers list at");
```
nuno-agostinho, line 268: This will not work as expected unless you put parentheses to `grep`:

```suggestion
            push @developer, $1 unless (grep(/^$1$/, @developer) || $1 eq "developers list at");
```

nuno-agostinho (changes requested):

nuno-agostinho (approved):
Working as expected, thanks @nakib103!

nuno-agostinho:
Merged to `release/112` and `main`.

PR #1059, opened 2023-11-30
Web VEP: fix incorrect variant end

nuno-agostinho:
[User-requested bug fix](https://genomes-ebi.slack.com/archives/C0JUVJV6W/p1701168853615039): This PR fixes the incorrect `END` location for short variants in web VEP by storing this information into the `INFO/END` field by default, as [already done for SVs](https://github.com/Ensembl/ensembl-variation/blob/4d3649a2a269a3d3ce49bf21f1f9a660d90dc533/modules/Bio/EnsEMBL/Variation/StructuralVariationFeature.pm#L1169).

## Motivation

Web VEP is returning incorrect `END` location for multiple variants when using rsIDs. Given that web VEP output is in VCF format, no `END` location is available by default. To prepare the variant location, web VEP first checks if it's stored in `INFO/END` field, otherwise it calculates based on the length of the reference allele: https://github.com/Ensembl/public-plugins/blob/main/tools/modules/EnsEMBL/Web/TmpFile/VcfTabix.pm#L254-L264

VEP stores the correct `END` position in the VariantFeature object. As such, we simply need to save this into the `INFO/END` field of the VCF output, [like in SVs](https://github.com/Ensembl/ensembl-variation/blob/4d3649a2a269a3d3ce49bf21f1f9a660d90dc533/modules/Bio/EnsEMBL/Variation/StructuralVariationFeature.pm#L1169). As this is done for SVs by default, this could also be the default for short variants.

## Testing

Compare results for the following variants between:
- VEP CLI with default output: correct location
- Web VEP in [current website](https://www.ensembl.org/Tools/VEP): calculated `END` location that is incorrect when compared with VEP CLI
- Web VEP in [my sandbox](http://wp-np2-11.ebi.ac.uk:6070/Tools/VEP): should have the same (correct) location as VEP CLI

```
rs746071566
rs3064744
rs72549309
rs267608279
rs72549346
rs72549356
rs72549352
rs72549354
rs72549353
rs5030656
```

### DIFF
--- a/modules/t/variationFeature.t
+++ b/modules/t/variationFeature.t
@@ -161,14 +161,14 @@ my $sr_start = $vf->seq_region_start;
 
 is_deeply(
   $vf->to_VCF_record(),
-  [$chr, $sr_start, $vname, 'A', 'T', '.', '.', '.'],
+  [$chr, $sr_start, $vname, 'A', 'T', '.', '.', "END=$end"],
   'to_VCF_record'
 );
 
 $vf->strand(-1);
 is_deeply(
   $vf->to_VCF_record(),
-  [$chr, $sr_start, $vname, 'T', 'A', '.', '.', '.'],
+  [$chr, $sr_start, $vname, 'T', 'A', '.', '.', "END=$end"],
   'to_VCF_record - rev strand'
 );
 $vf->strand($strand);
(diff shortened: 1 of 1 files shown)


nakib103:
The first variant got position - 
`22:42128250-42128254` - which kind of makes sense if we think as VCF representation

The second variant got - 
`22:42128174-42128178` - which is not a VCF representation, but can be considered correct if we think the allele is minimised.

nuno-agostinho:
Hey @nakib, do you think we should support have a `START` field, similar to `END`?

`START` could be created based on `$vf->{original_start}` (42128174 for rs5030656 and 42128249 for 42128249). This makes me question if we should also use `$vf->{original_end}` for the `END` location, what do you think?

nakib103 (changes requested):
Hi @nuno-agostinho,

Yes it makes sense to use `$vf->{original_start}`. We have `UPLOADED_ALLELE` and `REF_ALLELE` in the web VEP output. Than it would at least be matched with the `UPLOADED_ALLELE`. Furthermore this is also what is reported by the **Variant Table** page for location.

Currently it is not matching either of the allele from those two column values.

PR #1061, opened 2024-01-12
Bugfix - wrong HGVSp when `--shift_hgvs` is enabled

nakib103:
[ENSVAR-6131](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-6131)

In https://github.com/Ensembl/ensembl-variation/pull/994 we added a update that will convert a `delins` to `dup`. An example can be found in the issue reported in https://github.com/Ensembl/ensembl-vep/issues/500 with example (`chr1  20978294  .  TTTTATTTATTTA  TTTTATTTATTTATTTA`). The code change was in `_clip_allele` function which is used to check after clipping alleles the type of HGVS notation changes. We introduced checking whether type changes to `dup` before checking for `ins`.

The `_clip_allele` function is used by both HGVSc and HGVSp. For HGVSp this change caused a problem as the `dup` calculation in done after `_clip_allele` - in `_check_for_peptide_duplication` function, whereas, for HGVSc it is done before. 

So in these code changes I have added condition to not revert `delins` to `dup` for HGVSp under `_clip_allele`.

- also revert back checking `=` type for only HGVSp (updated in the above mentioned PR without any issue being reported)
- update some comments and refactoring of codes

### Test
HGVSp position -
```
#CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO	FORMAT	H_NJ-HCC1395-HCC1395
6	41754573	.	C	CCTT	.	.	.	GT:DP:DP4:BQ:SS:AD:DP2:TAR:TIR:TOR:DP50:FDP50:SUBDP50:FT	0/1:76:13,9,28,26:36:2:12:78:12,12:55,55:12,12:80.52:1.92:0:PASS
```

HGVSc/HGVSp `dup` gets corrected still -
```
chr1 20978294 . TTTTATTTATTTA TTTTATTTATTTATTTA
chr1 20978964 . GGTAGGC GGTAGGCGTAGGC
```

### DIFF
--- a/modules/Bio/EnsEMBL/Variation/TranscriptVariationAllele.pm
+++ b/modules/Bio/EnsEMBL/Variation/TranscriptVariationAllele.pm
@@ -1732,6 +1732,7 @@ sub hgvs_protein {
   $convert_to_three_letter = 1 unless defined $convert_to_three_letter;
   ##### Convert ref & alt peptides taking into account HGVS rules
   $hgvs_notation = $self->_get_hgvs_peptides($hgvs_notation, $convert_to_three_letter);
+
   unless($hgvs_notation) {
     $self->{hgvs_protein} = undef;
     return undef;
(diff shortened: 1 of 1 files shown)


likhitha-surapaneni (approved):
Tested and LGTM. Thanks @nakib103

nakib103:
merged to `main` and `postrealeasefix/112`

PR #1069, opened 2024-02-12
VEP plugins: retrieve data from specific Tabix-indexed files

nuno-agostinho:
ENSVAR-2596

Required by https://github.com/Ensembl/VEP_plugins/pull/681

## Motivation

Currently, Tabix-indexed data in VEP plugins when using `get_data` is retrieved in conglomerate from all the files.

This update allows to retrieve data only from a specific file by changing how the cache mechanism is created when `get_data` gets a file:
- Nothing should change if the last argument of `get_data` is not defined, so this change should not affect any plugins
- If `get_data` defines a file:
    - Result cache are separated by filename
    - Data is only retrieved for the specified filename

## Testing

1. Test with https://github.com/Ensembl/VEP_plugins/pull/681
2. Test with any plugin that uses `get_data` function to see if its results are as expected (such as `CADD` plugin).

### DIFF
--- a/modules/Bio/EnsEMBL/Variation/Utils/BaseVepTabixPlugin.pm
+++ b/modules/Bio/EnsEMBL/Variation/Utils/BaseVepTabixPlugin.pm
@@ -339,6 +339,7 @@ sub cache_size {
   Arg[1]     : chromosome
   Arg[2]     : start
   Arg[3]     : end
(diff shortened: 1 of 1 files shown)


jamie-m-a (approved):
Original functionality retained, new function added to support selection of specific input

PR #1075, opened 2024-03-21
Remove redundant symbolic alt to so term hash

nakib103:
Reported in https://github.com/Ensembl/VEP_plugins/issues/710
Related https://github.com/Ensembl/ensembl-vep/pull/1643

We have hash to convert from symbolic alternative allele from VCF to SO term in several places. It is best have it in one place to avoid errors.


modules/Bio/EnsEMBL/Variation/Utils/Config.pm
```diff
@@ -1382,5 +1383,28 @@ our %SO_ACC_MAPPER = (
...
+        INS_LINE1 => 'LINE1_insertion',
+        INS_SVA   => 'SVA_insertion',
+
+        DEL       => 'deletion',
+        DEL_ME    => 'mobile_element_deletion',
+        DEL_ALU   => 'Alu_deletion',
+        DEL_HERV  => 'HERV_deletion',
+        DEL_LINE1 => 'LINE1_deletion',
+        DEL_SVA   => 'SVA_deletion',
+
```
nuno-agostinho, line 1401: Add missing item:

```suggestion

TREP => 'tandem_repeat',
```

modules/Bio/EnsEMBL/Variation/Utils/VEP.pm
```diff
@@ -933,12 +924,7 @@ sub convert_to_vcf {
     else {
 
         # convert to SO term
-        my %terms = (
-            'insertion' => 'INS',
-            'deletion' => 'DEL',
-            'tandem_duplication' => 'TDUP',
-            'duplication' => 'DUP'
-        );
+        my %terms = reverse %SV_SO_TERMS;
```
nuno-agostinho, line 927: The ALT format in VCF format does not allow some of the items from `%terms` (according to [VCF 4.4 specs](https://samtools.github.io/hts-specs/VCFv4.4.pdf)). VCF only allows SV symbolic alleles in ALT starting with the following terms:
* DEL
* INS
* DUP
* INV
* CNV

Instead of modifying `%terms` (may have potential issues?), we could add logic here to simply convert:
- `TDUP` to `DUP:TANDEM`
- `TREP` to `CNV:TR`
- `BND` to the alternative allele itself
- `INS_ALU` to `INS:ME:ALU` (and the same for the other cases for mobile element insertion/deletion); maybe it would be easier to add a `ME_` to the key of all mobile elements and replace `_` with `:`?

modules/Bio/EnsEMBL/Variation/Utils/VEP.pm
```diff
@@ -478,12 +479,7 @@ sub parse_ensembl {
         my $so_term;
 
         # convert to SO term
-        my %terms = (
-            INS  => 'insertion',
-            DEL  => 'deletion',
-            TDUP => 'tandem_duplication',
-            DUP  => 'duplication'
-        );
+        my %terms = %SV_SO_TERMS;
```
nuno-agostinho, line 482: A SV based on `BND` here won't include the breakends. Do we want to support the `BND` here?

modules/Bio/EnsEMBL/Variation/Utils/VEP.pm
```diff
@@ -631,12 +627,7 @@ sub parse_vcf {
 
         if(defined($type)) {
             # convert to SO term
-            my %terms = (
-                INS  => 'insertion',
-                DEL  => 'deletion',
-                TDUP => 'tandem_duplication',
-                DUP  => 'duplication'
-            );
+            my %terms = %SV_SO_TERMS;
```
nuno-agostinho, line 630: The keys in `%SV_SO_TERMS` are not the proper VCF types. We would need to run `get_SO_term($type)` to get the correct SO term.

nuno-agostinho (changes requested):
Hi @nakib103, thanks for your PR! I put some concerns regarding the changes in the following comments.

Please tell me if you want to discuss how to best approach my suggestions. Thanks!

nakib103:
@nuno-agostinho, 
The unit test on [one before the last](https://github.com/Ensembl/ensembl-variation/pull/1075/commits/d252984af4cc3c2302e8c445e02a26b3990a966e) take accounts the change in ensembl-variation and passed. You can ignore the unit test failure in the last commit.

modules/Bio/EnsEMBL/Variation/Utils/VEP.pm
nakib103: I tested with a variant like this - `1 7936271 7936271 BND + bnd` and from `parse_ensembl` got -
```
{
                 'class_SO_term' => 'chromosome_breakpoint',
                 'start' => '7936271',
                 'adaptor' => undef,
                 'variation_name' => 'bnd',
                 'chr' => '1',
                 'strand' => 1,
                 'end' => '7936271'
               }, 'Bio::EnsEMBL::Variation::StructuralVariationFeature' );
```

seems to work.

nuno-agostinho (changes requested):
Thanks for your PR, @nakib103! It's working as intended when testing `StructuralVariantOverlap` with `same_type=1`.

I am going to approve it, but I am concerned about moving some functions to ensembl-vep, as it makes ensembl-variation require VEP to properly test and may break other repos (web? REST?) that did not consider this requirement before.

Please check with the team if this if fine. If so, the PR is ready to be merged.

modules/Bio/EnsEMBL/Variation/Utils/VEP.pm
```diff
@@ -933,14 +859,16 @@ sub convert_to_vcf {
...
-            'deletion' => 'DEL',
-            'tandem_duplication' => 'TDUP',
-            'duplication' => 'DUP'
-        );
+        my %terms = reverse %SO_TERMS;
+        my $abbrev = $terms{$vf->class_SO_term} || $vf->class_SO_term;
+
+        $abbrev = "DUP:TANDEM" if $abbrev eq "TDUP";
+        $abbrev = "CNV:TR" if $abbrev eq "TREP";
+        $abbrev =~ s/_/:/ if $abbrev =~ /^(INS|DEL)_ME$/;
```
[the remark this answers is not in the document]
nakib103: Noted! but I think that requires a bigger change and not within the scope/objective of this PR.

nakib103:
Closing this PR and will create a new one.

PR #1077, opened 2024-04-10
Fix HGVSg for multiple alternative alleles

nuno-agostinho:
Fixes Ensembl/ensembl-vep#1653

When using multiple alternative alleles with the `--hgvsg` option in VEP, the HGVSg is incorrect and different compared to running VEP on each alternative allele separately.

The bug occurs because we check the variation class SO term based on the complete allele string. In the case of an allele string with multiple alternative alleles (e.g., `GG/-/G`), we need to check the class SO term for each alternative allele separately.

## Testing

Run VEP on GRCh37 with `--hgvsg`. Example variants:

```
chr8 143694490 143694490 T/TT/TTT +
chr8 143694490 143694492 GGG/GG/G +
chr14 103576804 103576805 GG/-/G +
chr14 103576804 103576805 GG/- +
chr14 103576804 103576805 GG/G +
```

The HGVSg for each alternative allele should be:
- `GG/-`: `chr14:g.103576811_103576812del` 
- `GG/G`: `chr14:g.103576812del`

### DIFF
--- a/modules/Bio/EnsEMBL/Variation/VariationFeature.pm
+++ b/modules/Bio/EnsEMBL/Variation/VariationFeature.pm
@@ -1121,6 +1121,31 @@ sub class_SO_term {
     return $self->{class_SO_term};
 }
 
+=head2 var_allele_class
+
+    Args[1]     : allele
+    Example     : my $variation_allele_class = $vf->var_allele_class('AA/A');
+    Description : returns the Ensembl term for the class of this variation allele
+    ReturnType  : String
+    Exceptions  : throws if we can't find a corresponding display term for an SO term
+    Caller      : General
+    Status      : Stable
+
+=cut
+
+sub var_allele_class {
+    my $self    = shift;
+    my $allele  = shift;
+
+    my $so_term = SO_variation_class($allele);
+
+    # convert the SO term to the ensembl display term
+    my $class = $self->is_somatic ?
+        $VARIATION_CLASSES{$so_term}->{somatic_display_term} :
+        $VARIATION_CLASSES{$so_term}->{display_term};
+    return $class;
+}
+
 =head2 get_all_evidence_values
 
   Arg [1]     : none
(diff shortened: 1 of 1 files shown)


likhitha-surapaneni (changes requested):
Hi @nuno-agostinho , thank you for the PR.

According to the 3' rule, shouldn't the HGVSg for alternative alleles be:

- GG/-: chr14:g.103576811_103576812del
- GG/G: chr14:g.103576812del

modules/Bio/EnsEMBL/Variation/VariationFeature.pm
```diff
@@ -1934,7 +1961,14 @@ sub hgvs_genomic {
     ### Apply HGVS 3' shift if required
     my $offset = 0;
     my $lookup_order = 1;
-    my $var_class  =  $self->var_class();
+
+    if ($is_multi_allelic) {
+      # fix for multi-allelic variants
+      ($ref_allele, $allele, $chr_start, $chr_end, my $change) = @{trim_sequences($ref_allele, $allele, $chr_start)};
+      $ref_start += $change;
```
likhitha-surapaneni, line 1968: `$change` is a boolean value, so it may not capture how many bases we want `$ref_start` shifted by. 
Are we interested in 
```
my $old_chr_start = $chr_start;
($ref_allele, $allele, $chr_start, $chr_end, my $change) = @{trim_sequences($ref_allele, $allele, $chr_start)};
my $offset = $chr_start - $old_chr_start;
$ref_start += $offset;
```

Example to test: `chr8    143694490   143694492   GGG/GG/G    +   var_test`

nuno-agostinho:
Thanks for the review, @likhitha-surapaneni. Also fixed an error when using insertions using multiple ALT alleles, such as:
```chr8 143694490 143694490 T/TT/TTT +```

likhitha-surapaneni (approved):
Thanks for the changes @nuno-agostinho , the PR is working as expected

PR #1079, opened 2024-04-15
Add option to ClinVar to populate transcript variation

dglemos:
Add option `-insert_tv` to populate transcript_variation and MTMP_transcript_variation.
By default, the script will not import consequences into those two tables (original behaviour).

Ticket: [ENSVAR-6285](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-6285)

### DIFF
--- a/scripts/import/import_clinvar_xml
+++ b/scripts/import/import_clinvar_xml
@@ -1380,9 +1380,6 @@ sub enter_var{
 
   $var_feat_adaptor->store($vf);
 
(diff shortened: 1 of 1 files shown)


jamie-m-a (approved):
Going to approve as the old functionality is restored, but as noted in JIra, the TV update mode is not currently working as intended and will need work ahead of e114

dglemos:
The functionality has been tested again and it's working as expected.

PR #1080, opened 2024-04-17
updating for gencode primary

jamie-m-a:
Updating variation consequence pipeline to have an option to run on only gencode primary transcripts.

Also improved the debug method to make it more useful and switched on from config.

Full details on ticket: ENSVAR-6238


nuno-agostinho (approved):
Thanks for the changes, @jamie-m-a!

New options `debug_genes` and `gencode_primary` seem to be working as intended.

PR #1087, opened 2024-05-09
Fix less and more than symbols in usage examples

nuno-agostinho:
Related with https://github.com/Ensembl/public-plugins/pull/790


likhitha-surapaneni (approved):
Works as expected. Thanks @nuno-agostinho

PR #1093, opened 2024-05-28
Updating cadd and dbNSFP paths

jamie-m-a:
Updating CADD and dbNSFP files to latest versions.


likhitha-surapaneni (approved):
LGTM, thanks @jamie-m-a

PR #1105, opened 2024-07-08
Update ClinVar format part 1

dglemos:
Ticket: https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-5941

This PR includes:
- update to new format: `/ReleaseSet/ClinVarSet/ReferenceClinVarAssertion/ClinicalSignificance` updated to `/ReleaseSet/ClinVarSet/ReferenceClinVarAssertion/Classifications/GermlineClassification`
- skip when variant has clinical significance "no classifications from unflagged records"
- remove method to extract extra citations
- fix code for conflicting clinical significance

Next step is to support somatic clinical significance which is going to be done in a separate PR.

### DIFF
--- a/scripts/import/import_clinvar_xml
+++ b/scripts/import/import_clinvar_xml
@@ -230,13 +230,17 @@ sub process_clinvar_set {
     }
 
     ## clinical significance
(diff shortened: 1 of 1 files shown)


nakib103 (approved):
LOTM!~

PR #1109, opened 2024-07-23
Nextflow: print instructions if JVM memory is low

nuno-agostinho:
* Throw error if memory used for Nextflow head node is lower than a minimum value
    - JVM memory is 1/4 of the SLURM job memory by default
    - The default minimum value of 0.4 GiB is a bit lower than 1/4 of a 2 GiB SLURM job
    - The error instructions show easy solutions for the issue
* Uniformise all pipelines to use common functions

### DIFF
--- a/nextflow/utils/utils.nf
+++ b/nextflow/utils/utils.nf
@@ -1,20 +1,20 @@
 #!/usr/bin/env nextflow
 
-def print_params(description, separator="-"*description.length(), nullable=true,
-                 indent=4, params=params, sort=false, skip=['help']) {
+def print_params(description, nullable=[], sort=false, skip=['help'], 
+                 params=params, separator="-"*description.length(), indent=4) {
   /*
     Print script parameters
 
     @param description Short pipeline description
+    @param nullable    List of params allowed to be null; raises an error for
+                       null params that are not in this list (default: [])
+    @param sort        Boolean to sort params alphabetically (default: false)
+    @param skip        List of params not to print; if null, all params are
+                       printed (default: ['help'])
+    @param params      Map of params (default: script params)
     @param separator   Separator used between description and params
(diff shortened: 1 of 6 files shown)


nextflow/utils/utils.nf
```diff
@@ -60,3 +60,25 @@ def print_summary() {
...
+    @param min Min memory in GiB (default: 0.4)
+  */
+
+  mem = Runtime.getRuntime().maxMemory() / (1024 ** 3) // in GiB
+  if (mem < min) {
+    log.error """
+    ERROR: The memory of the Nextflow head node (${mem.round(2)} GiB) is below the recommended (${min} GiB); please try the following actions:
+      - Increase the JVM's max RAM percentage: export NXF_JVM_ARGS="-XX:InitialRAMPercentage=25 -XX:MaxRAMPercentage=75"
+      - Increase the JVM heap size memory: export NXF_OPTS="-Xms50m -Xmx${min}g"
+      - Increase the memory of the job used to run the Nextflow pipeline: salloc --time 6:00:00 --mem ${min * 4}GB
```
likhitha-surapaneni, line 80: salloc --mem seems to only take whole numbers

likhitha-surapaneni: There is also a class for [memory unit](https://www.nextflow.io/docs/latest/script.html#memoryunit):

nuno-agostinho: Unfortunately, the memory unit class automatically rounds down the values, so I don't think it is useful in this context (if only it would round up...)

I rounded up the memory values in the commands to work as expected. Thanks @likhitha-surapaneni!

likhitha-surapaneni (approved):
LGTM and working as expected, thanks @nuno-agostinho

PR #1110, opened 2024-07-31
Fix format checking for variant at start of a seq region

nakib103:
If we have a variant described at the start of a sequence it will fail for default VEP input format. For example, for a variant -

```
HSCHR15_4_CTG8	1	0	-/TC	+	var
```

The line `$line[2] && $line[2] =~ /^\d+$/ &&` will fail because the end position is `0` which is interpreted as False. We should use `exists` function instead to check if we have a field present.


nuno-agostinho (approved):
Thanks @nakib103, makes sense and it's working as expected!

nuno-agostinho:
Merged to `main` and `release/113`.

PR #1113, opened 2024-08-06
GO plugin: avoid GO file dump `match` warnings

nuno-agostinho:
The GO plugin is printing warnings related with the `match` parameter not being defined when running the GO dumps script.

This PR fixes the issue.


likhitha-surapaneni (approved):
LGTM, thanks @nuno-agostinho

PR #1122, opened 2024-09-05
Update release version in travis & README (e114)

nakib103:


dglemos (approved):
The failured is caused by ensembl-funcgen missing release/114.

PR #1123, opened 2024-09-13
EVAImport - allow csi index

nakib103:
- Allow CSI index for input VCF file
- Fix issue with `indent()` method not found (used `collect{}` to indent string)
- Update help text with mandatory `--version` parameter


jamie-m-a (approved):
All looks good to me - thanks Nakib

PR #1124, opened 2024-09-26
Fix hgvsg for insertion with ref allele `-`

nakib103:
### Problem

Here https://github.com/Ensembl/ensembl-variation/pull/1077 we have fixed hgvsg for multi-allelic variant. But it introduced a bug for insertion types.

We are not sending the end position during trimming -
https://github.com/nuno-agostinho/ensembl-variation/blob/3ed1013c5bfacb2416db9e67d50cb96fb2fcb0b6/modules/Bio/EnsEMBL/Variation/VariationFeature.pm#L1970

What it does for insertion is that it increase the end position by `length of reference sequence -1` - 
https://github.com/nuno-agostinho/ensembl-variation/blob/3ed1013c5bfacb2416db9e67d50cb96fb2fcb0b6/modules/Bio/EnsEMBL/Variation/Utils/Sequence.pm#L970

This becomes problematic if the `length of reference sequence -1` is not the same as `end - start` and there is no trimming happening so that position sticks around. This happen only in the case when we have `-` in ref allele for insertion. 

The end result is different position in slice and sub slice -
https://github.com/nuno-agostinho/ensembl-variation/blob/3ed1013c5bfacb2416db9e67d50cb96fb2fcb0b6/modules/Bio/EnsEMBL/Variation/VariationFeature.pm#L2016-L2019

and, terminating in a error here -
https://github.com/nuno-agostinho/ensembl-variation/blob/3ed1013c5bfacb2416db9e67d50cb96fb2fcb0b6/modules/Bio/EnsEMBL/Variation/Utils/Sequence.pm#L511

### Fix
- Simply provide end position so `trim_sequences` do not try to calculate it.
- And, update the `ref_end` position. Otherwise it will fail in trim on reverse direction.

### Test:

```
chr8 143694491 143694490 -/TT/TTT +
chr8 143694491 143694490 -/GG/GGG +
chr8 143694490 143694492 GTT/GT/T + (trim on reverse direction)
```

### DIFF
--- a/modules/Bio/EnsEMBL/Variation/VariationFeature.pm
+++ b/modules/Bio/EnsEMBL/Variation/VariationFeature.pm
@@ -1967,7 +1967,7 @@ sub hgvs_genomic {
     if ($is_multi_allelic) {
       # fix for multi-allelic variants
       my $old_chr_start = $chr_start;
-      ($ref_allele, $allele, $chr_start, $chr_end) = @{trim_sequences($original_ref_allele, $allele, $chr_start)};
+      ($ref_allele, $allele, $chr_start, $chr_end) = @{trim_sequences($original_ref_allele, $allele, $chr_start, $chr_end)};
       $allele ||= '-';
       $offset = $chr_start - $old_chr_start;
       $ref_start = $original_ref_start + $offset;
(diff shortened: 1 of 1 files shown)


olaaustine (approved):
LGTM

olaaustine (approved):

olaaustine:
Merged into main and release/114 also

PR #1129, opened 2024-10-31
Update regulation pipeline

dglemos:
Add long queue to finish_regulation_effect


nakib103 (approved):
Needed for GRCh37 in 114, should be default for all cases in future.
LGTM!~

PR #1130, opened 2024-11-08
optimising for human

jamie-m-a:
I've manually set these in eHive for running the pipeline in 113 and 114 with no problems.

It should not have an impact (other than speeding up) for other species, but I would recommend to test it out on mouse and confirm the transcript variation / variation feature tables are generated as expected (compare to a handed over DB).


nakib103 (approved):
LGTM !~

PR #1134, opened 2024-11-19
Update transcript variation pipeline - remove ehive version

dglemos:
It was announced the 2.6 branch doesn't have Slurm support (officially) and is now behind.
The new version start at 2.7.0


nakib103:
merged to `main` and `release/114`

PR #1137, opened 2024-12-06
Remove redundant SO term hash

nakib103:
Simpler version of https://github.com/Ensembl/ensembl-variation/pull/1075
Remove function swapping with ensembl-vep.


nakib103:
merged to `main` and `release/114`

PR #1141, opened 2025-01-02
Fix flat file genotype/population calculation

nakib103:
- `utils.nf` 
  Have a common place for all EVA related function. As we were replicating them in multiple script it was getting cumbersome to update.
  - `sub get_vcf_content_types`: 
    - Improve checking of `population` type - query database to check `sub is_freq_from_gts` is set.
    - Fix checking of `genotype` type - `sub genotype_samples_exists` was missing some cases where sample name had population prefix from database and not vcf_config.json.

- [Species list](http://wp-np2-11.ebi.ac.uk:7070/info/genome/variation/species/species_data_types.html)
  - (species_list.pl) Fixed missing genotype ticks for some database species.
  - (species_list_vcf.pl) Added population column for VCF species (before it was only genotypes).
  - (species_list_vcf.pl) Fixed count difference in the VCF species table (before they were missing).

- [Species detailed list](http://wp-np2-11.ebi.ac.uk:7070/info/genome/variation/species/species_detailed_counts.html) (TBD)
  - Add species with population data (American mink and Nile tilapia).
  - Instead of having same count for population and genotype, irrespective of what data the VCF file have, we now have the count in appropriate column and leave the other column blank if applicable.

- [Sources list ](http://wp-np2-11.ebi.ac.uk:7070/info/genome/variation/species/sources_documentation.html)
  - Fix mismatch between "Species list" (top-right) and actual table of the species; by preventing adding hidden empty row - https://github.com/Ensembl/ensembl-variation/blob/4658db7531c41db555a0a3384a3b3cc3308e03aa/scripts/docs/sources2html.pl#L580
  - Fix no color change in alternating rows in EVA species
  - For new version of EVA script now support adding info in "Species list" (top-right) and color change in table.

### DIFF
--- a/scripts/docs/update_web_documentation.pl
+++ b/scripts/docs/update_web_documentation.pl
@@ -344,7 +344,7 @@ if ( $sections eq "all" || grep(/^${section}$/, @sections) ){
   $tmp_file  = $file_name;
 
   print localtime() . "\t# Start sources list ...\n";
-  `perl $dirname/sources2html.pl -v $version -o $tmp_file -hlist $hlist -phost $phost -config $vcf_config_file -d_dir $data_dir`;
+  `perl $dirname/sources2html.pl -v $version -o $tmp_file -hlist $hlist -phost $phost -config $vcf_config_file -d_dir $data_dir -p_data $p_data`;
 
   $copy2subdir = ($no_subdir) ? '' : $subdirs{$file_name};
   copy_updated_file($copy2subdir,$file_name,$tmp_file);


jamie-m-a (commented):
I'm going to assume this is an improvement as the current set up is messy and repetitive as you say.

One minor thing it would be nice to have it highlighted in the table (VCF species) when a species is new. So for example for Domestic yak, it would say 30 million, then the second column could say "NEW for eXXX", rather than just +30 million

nakib103:
Hi @jamie-m-a , created the ticket [ENSVAR-6689](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-6689) with the requirement.

PR #1145, opened 2025-01-15
Github actions creation to update copyright

olaaustine:


.github/workflows/update_copyright.yml
```diff
@@ -0,0 +1,51 @@
...
+        with: 
+          new_branch: ${{ env.branch_name }}
+          add: '.'
+          message: "Update copyright year"
+
+      - name : Create pull request
+        env: 
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: | 
+          gh pr create -B main -H $branch_name --title "Update Copyright year" --body 'Created by Github action'
```
likhitha-surapaneni, line 51: Here, base branch is `main` as opposed to `postreleasefix/xx` as this can be tricky to code.
One way is to take the latest `postreleasefix/xx` from the existing branches but not sure how easy that would be. This looks good for the first iteration.

likhitha-surapaneni (commented):
Thank you for this feature @olaaustine , this looks good. Can we try running a successful action in the forked branch with different cron job settings if possible?

.github/workflows/update_copyright.yml
olaaustine: Decided not to use postreleasefix/** as you already said difficult to code and also most times our main and postreleasefix/** have the same "logs" but yeah this is okay for the first iteration. We can make it better for the future.

olaaustine:
Note - 
Once this is merged,  need to make same PR to all of variation repos

PR #1147, opened 2025-01-23
Remove XML::Hash::XS from cpanfile

dglemos:
`XML::Hash::XS` breaks the tests for perl 5.14
As this package is only used by the ClinVar import we can temporarily remove it from cpanfile. Later we'll find another package to import XML and/or remove 5.14 from the tests.


likhitha-surapaneni (approved):
Tests passed. LGTM

likhitha-surapaneni:
Merged to `main` and `release/114`

PR #1149, opened 2025-01-31
ENSVAR-6654 & ENSVAR-6381

olaaustine:
#https://github.com/Ensembl/ensembl-vep/issues/1710
#https://github.com/Ensembl/ensembl-vep/issues/1796

### DIFF
--- a/modules/Bio/EnsEMBL/Variation/Utils/VariationEffect.pm
+++ b/modules/Bio/EnsEMBL/Variation/Utils/VariationEffect.pm
@@ -1594,4 +1594,7 @@ sub contains_entire_feature {
     my $bvf  = $vfo->base_variation_feature;
     my $feat = $vfo->feature;
 
-    return ( ($bvf->{start} <= $feat->{start}) && ($b
+    return ( ($bvf->{start} <= $feat->{start}) && ($bvf->{end} >= $feat->{end}) ); 
+}
+
+1;
--- a/modules/Bio/EnsEMBL/Variation/Utils/VariationEffect.pm
+++ b/modules/Bio/EnsEMBL/Variation/Utils/VariationEffect.pm
@@ -1346,10 +1346,10 @@ sub ref_eq_alt_sequence {
    
    my $final_stop_length = length($final_stop) if defined($final_stop) ne '';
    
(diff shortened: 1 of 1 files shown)


jamie-m-a:
Hi Ola, I don't think we need to roll this back completely, it was added to address the situations described in: https://github.com/Ensembl/ensembl-variation/pull/1066 

I don;t agree with all of Nuno's use case examples, but the stop retained examples of * to * in the same position should be captured and can be with something like: 

`($ref_pep eq substr($alt_pep, 0, 1) && $alt_pep =~ /\*/ && $ref_pep =~/\*/  && pos($alt_pep) == pos($ref_pep))`

I tested it on the GRCh37 examples linked in this ticket and it correctly returns stop_gained as the top consequence. It also returned stop_retained for the first example Nuno gave (6	30590698	.	GTGA	CTAG on GRCh38).

Could you check it out a bit more on some more tests - thanks

olaaustine:
> Hi Ola, I don't think we need to roll this back completely, it was added to address the situations described in: #1066
> 
> I don;t agree with all of Nuno's use case examples, but the stop retained examples of * to * in the same position should be captured and can be with something like:
> 
> `($ref_pep eq substr($alt_pep, 0, 1) && $alt_pep =~ /\*/ && $ref_pep =~/\*/ && pos($alt_pep) == pos($ref_pep))`
> 
> I tested it on the GRCh37 examples linked in this ticket and it correctly returns stop_gained as the top consequence. It also returned stop_retained for the first example Nuno gave (6 30590698 . GTGA CTAG on GRCh38).
> 
> Could you check it out a bit more on some more tests - thanks

Hi Jamie, 
With more tests, this change will not capture all examples of stop retained especially with the [first issue ]((https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-4348)), I am going to keep investigating to figure out the best possible way around this. 
I do not want to hold up the branching so we can always rebase as already discussed during standup.
Thanks

jamie-m-a:
Agreed Ola, we shouldn’t rush this change, thanks for taking a look.

jamie-m-a:
Can rebase now

jamie-m-a:
Clsoing this one down, it's been superseded by https://github.com/Ensembl/ensembl-variation/pull/1184

PR #1150, opened 2025-02-07
patching test DBs

jamie-m-a:
Patching script ran successfully, bit note that variation has already been patched earlier by Diana for testing the ClinVar import, hence not present in this patch.


modules/t/test-genome-DBs/multi/ontology/meta.txt
```diff
@@ -59,3 +59,4 @@
 62	patch	patch_112_113_a.sql|schema_version	\N
 63	patch	patch_112_113_b.sql|Ensure meta_value is not null	\N
 64	patch	patch_113_114_a.sql|schema_version	\N
+65	patch	patch_114_115_a.sql|schema_version	\N
```
likhitha-surapaneni, line 62: Hi @jamie-m-a , do we want to add the same changes to core, variation and otherfeatures?

- `modules/t/test-genome-DBs/homo_sapiens/core/meta.txt`
- `modules/t/test-genome-DBs/homo_sapiens/variation/meta.txt`
- `modules/t/test-genome-DBs/homo_sapiens/otherfeatures/meta.txt`

jamie-m-a: Technically I guess, they don't seem to get done by the script, so I imagine they're not actually important for our tests. But I'll add them for consistency

jamie-m-a: Odd. Will add.

jamie-m-a: Variation is already present as per Diana's changes, I've added for core and otherfeatures.

modules/t/test-genome-DBs/homo_sapiens/core/meta.txt
```diff
@@ -259,3 +259,4 @@
 1059	\N	patch	patch_112_113_a.sql|schema_version
 1060	\N	patch	patch_112_113_b.sql|Ensure meta_value is not null
 1061	\N	patch	patch_113_114_a.sql|schema_version
+1062    \N	patch   patch_114_115_a.sql|schema_version
```
likhitha-surapaneni, line 262: Thanks @jamie-m-a , there are just some final changes for upgrading schema_version to 115 and updating `table.sql` for core and otherfeatures

jamie-m-a: Sorry missed the schema versions...! Updated now. There's no need to increment ISAM for meta, they seem to already be at the correct numbers for core and otherfeatures (I had checked that :) )

PR #1154, opened 2025-03-13
Update hgmd script - set somatic to 0 when creating variation_feature

dglemos:
Ticket: [ENSVAR-6688](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-6688)


jamie-m-a (approved):
Looks good - thanks Diana

PR #1155, opened 2025-03-17
Phenotype pipeline: consider non reference species

nakib103:
[ENSVAR-6663](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-6663)

### Import of non reference species and update sheep
Add import for non reference species such as chicken red jungle fowl and cat abyssinian for OMIA and AnimalQTL
**AnimalQTL**: Add non reference species to the list and add their respective url for download
**OMIA**: Add non reference species to the list and copy the file for reference species as was done for sheep and dog before. Made the logic more generic by adding a hash `%nonref_species` to list all such species.

For sheep we have now changed according to the reference update. So, `ovis_aries` is now rambouillet and `ovis_aries_texel` is for texel.

### Separate input file and path for AnimalQTL
It seemed we were setting `inputFile` as absolute where `animalqtl_inputDir` does not exist or the downloaded file not exist in that directory. But where the file exist we were taking the relative path. But subsequent logic assumed relative path ([e.g.](https://github.com/Ensembl/ensembl-variation/blob/ebfce7401437567284b8d6b4e8bbcdb0cc64b898/modules/Bio/EnsEMBL/Variation/Pipeline/PhenotypeAnnotation/ImportAnimalQTL.pm#L189)) and would cause error when the first 2 cases happen.

We now separate `inputFilePath` which is absolute path from `inputFile` which is relative and use them appropriately

### Other fixes
- The `run_cmd` do not return any value so stopped expecting it.
- For AnimalQTL when we look for existing file in `animalqtl_inputDir` we now look for exact file instead of `/^QTL_$species.*gff3$/` because it would match `ovis_aries_texel` when we run for `ovis_aries`.

Test pipeline: http://guihive.ebi.ac.uk:8080/versions/96/?driver=mysql&username=ensadmin&host=mysql-ens-var-prod-4&port=4694&dbname=snhossain_ehive_others_phenotype_annotation_115&passwd=xxxxx

Felis_catus failed with decreased count which seems to be expected as `felis_catus_variation_115_1` which is the now the reference cat has less number of phenotype imported than previous reference - abyssinian.

### DIFF
--- a/modules/Bio/EnsEMBL/Variation/Pipeline/PhenotypeAnnotation/Constants.pm
+++ b/modules/Bio/EnsEMBL/Variation/Pipeline/PhenotypeAnnotation/Constants.pm
@@ -116,11 +116,11 @@ use constant SOURCES_IN_RUN_TYPES => ( OMIA      => 'ANIMALSET',
 use constant SPECIES => ( 'RGD'       => ['rattus_norvegicus'],
                           'ZFIN'      => ['danio_rerio'],
 
-                          'ANIMALQTL' => ['bos_taurus', 'gallus_gallus', 'equus_caballus',
-                                          'sus_scrofa', 'ovis_aries', 'ovis_aries_rambouillet', 'capra_hircus'],
-                          'OMIA'      => ['felis_catus','gallus_gallus','capra_hircus',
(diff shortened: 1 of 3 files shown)


dglemos (approved):

dglemos:
Merged into release/115 and main

PR #1156, opened 2025-03-19
Improve MaveDB nextflow pipeline 

ainefairbrother:
## Description

This PR is for major improvements to the MaveDB pipeline and the way it ingests MaveDB data. 

## Use case

These improvements allow files from MaveDB's database dump (stored on Zenodo) to be supplied directly to the pipeline rather than relying on the MaveDB API to pull down the data on-the-fly, as this method can be unstable.

## Benefits

- Improved stability due to removing reliance on the MaveDB API. 
- Decreased run time as files are supplied directly rather than pulled down from the MaveDB API. 

## Possible Drawbacks

- The pipeline remains reliant on the specific structure of the files supplied by MaveDB for correct parsing. This is particularly true for the metadata format coercion code . I aim to find time to improve this and rather than coerce, to work with the format directly in the script as has been done for the mappings JSON data. I think this would make de-bugging easier. 
- Added complexity to the pipeline - it now has --from-files 'true' and 'false' branches of functionality that operate in slightly different ways to achieve the same goal. 
- Changes to the surrounding code may have unintended consequences on the API route of the pipeline. 

## Testing

- Previously, the pipeline contained no unit tests. I have added additional edge case checks relevant to the new data, catches and logging to help track why specific URN IDs may fail. 
- The major test of data validity was to take the output of the new file download route and test it against the output of the old API route. The idea behind this test was that any entries that are present in both MaveDB versions* should be identical between the two output files. 
- The data passed this test. 

\* Note that the output files wouldn't be expected to be fully identical as the version of MaveDB differs between that utilised by the original iteration of this pipeline and this iteration (MaveDB has lots more data now).

### Test 1: old data vs. new data - is the new data consistent with equivalent entries in the old data?

Question: for the URN IDs that are consistent between v1 of the MaveDB data processed by the old API-based nextflow pipeline and v2 of the MaveDB data processed by this new version of the pipeline - are the same scores assigned?

I tested this using the following R script to compare the two tables:

```R
# import libs
library(tidyverse)
library(arsenal)
library(janitor)

# unzip each dataset
system("cp /nfs/production/flicek/ensembl/variation/data/MaveDB/MaveDB_variants.tsv.gz /hps/nobackup/flicek/ensembl/variation/fairbrot/MaveDB/MaveDB_variants_api.tsv.gz; gunzip /hps/nobackup/flicek/ensembl/variation/fairbrot/MaveDB/MaveDB_variants_api.tsv.gz")
system("cd /hps/nobackup/flicek/ensembl/variation/fairbrot/MaveDB; cp MaveDB_variants.tsv.gz MaveDB_variants_dl.tsv.gz; gunzip MaveDB_variants_dl.tsv.gz")

# import data generated by the previous version of the pipeline
api_file = vroom::vroom("/hps/nobackup/flicek/ensembl/variation/fairbrot/MaveDB/MaveDB_variants_api.tsv") %>% # size = 1198664
  janitor::clean_names() %>% 
  dplyr::select(number_chr:urn, pro, accession, score) %>% 
  dplyr::filter(number_chr=="2") %>% 
  dplyr::mutate(id = paste0(number_chr, "_", start, "_", end))

# import data generated by the new version of the pipeline
dl_file = vroom::vroom("/hps/nobackup/flicek/ensembl/variation/fairbrot/MaveDB/MaveDB_variants_dl.tsv") %>% # size = 7672968
  janitor::clean_names() %>% 
  dplyr::select(number_chr:urn, pro, accession, score) %>% 
  dplyr::filter(number_chr=="2") %>% 
  dplyr::mutate(id = paste0(number_chr, "_", start, "_", end))

# the version of the MaveDB data processed by the old and the new pipeline
# is different, so in order to compare, find the intersection of variants
# i.e. those in both MaveDB versions
shared_ids = intersect(api_file$id, dl_file$id)

# filter and sort both datasets for shared variant IDs
api_file = api_file %>% 
  dplyr::filter(id %in% shared_ids) %>% 
  dplyr::arrange(id)

dl_file = dl_file %>% 
  dplyr::filter(id %in% shared_ids) %>% 
  dplyr::arrange(id)

# map two tables together and check that the scores being assigned to 
test = api_file %>% 
  tibble::rowid_to_column("api_id") %>% 
  dplyr::mutate(api_id = paste0("api_", api_id)) %>% 
  dplyr::left_join(x=., 
                   y=., 
                   by=c("number_chr", "start", "end", "ref", "alt", "urn", "accession", "id")) %>% 
  dplyr::mutate(score_eq = case_when(
    score.x==score.y ~ TRUE, 
    TRUE ~ FALSE
  )) %>% 
  dplyr::mutate(hgvs_eq = case_when(
    hgvs.x==hgvs.y ~ TRUE, 
    TRUE ~ FALSE
  )) %>% 
  dplyr::mutate(pro_eq = case_when(
    (pro.x==pro.y) | (is.na(pro.x) & is.na(pro.y)) ~ TRUE, 
    TRUE ~ FALSE
  ))

# check that the only value for each test is TRUE, meaning that all rows in the 
# column match between the two datasets
test$score_eq %>% unique() # are the scores mapped to each variant the same?
test$hgvs_eq %>% unique() # are the hgvs IDs mapped to each variant the same?
test$pro_eq %>% unique() # are the protein IDs mapped to each variant the same?
```

The correct result is that each of the 3 tests at the bottom yield `TRUE`, meaning that in both datasets, the same variant is assigned the same HGVS ID, the same protein ID and the same MaveDB score. This means that the two versions of the pipeline generate the same data for the same set of variants. 

### Test 2: does the new data work with the MaveDB plugin?

For this, the PlatinumGenomes VCF will be used as input, and the number of variants annotated by the MaveDB plugin will be assessed. 

#### Test 2a. Run VEP with MaveDB plugin and new data

```bash

input_vcf="/nfs/production/flicek/ensembl/variation/data/PlatinumGenomes/NA12878.vcf.gz"

# test with new mavedb pipeline data
echo "Running VEP with MaveDB plugin..."
/hps/software/users/ensembl/variation/fairbrot/ensembl-vep/vep \
    --input_file $input_vcf \
    --output_file "/hps/nobackup/flicek/ensembl/variation/fairbrot/MaveDB/test_with_plugin/mavedb_new_pipeline_out.vcf" \
    --format vcf \
    --vcf \
    --offline \
    --cache \
    --cache_version 113 \
    --dir_cache /nfs/production/flicek/ensembl/variation/data/VEP/tabixconverted \
    --assembly GRCh38 \
    --fasta /nfs/production/flicek/ensembl/variation/data/Homo_sapiens.GRCh38.dna.toplevel.fa.gz \
    --verbose \
    --force_overwrite \
    --plugin MaveDB,file=/hps/nobackup/flicek/ensembl/variation/fairbrot/MaveDB/MaveDB_variants.tsv.gz

# count the number of MaveDB annotations
cd /hps/nobackup/flicek/ensembl/variation/fairbrot/MaveDB/test_with_plugin
grep -v '^#' mavedb_new_pipeline_out.vcf | grep 'urn:mavedb:' | wc -l
grep -v '^#' mavedb_new_pipeline_out.vcf | grep -o 'urn:mavedb:' | wc -l
```

This should give 45 lines where at least one URN ID is assigned, and 491 individual instances of URN IDs being assigned in the file. 

#### Test 2b. Run VEP with MaveDB plugin and old data

```bash
input_vcf="/nfs/production/flicek/ensembl/variation/data/PlatinumGenomes/NA12878.vcf.gz"

# test with new mavedb pipeline data
echo "Running VEP with MaveDB plugin (old pipeline data)"
/hps/software/users/ensembl/variation/fairbrot/ensembl-vep/vep \
    --input_file $input_vcf \
    --output_file "/hps/nobackup/flicek/ensembl/variation/fairbrot/MaveDB/test_with_plugin/mavedb_old_pipeline_out.vcf" \
    --format vcf \
    --vcf \
    --offline \
    --cache \
    --cache_version 113 \
    --dir_cache /nfs/production/flicek/ensembl/variation/data/VEP/tabixconverted \
    --assembly GRCh38 \
    --fasta /nfs/production/flicek/ensembl/variation/data/Homo_sapiens.GRCh38.dna.toplevel.fa.gz \
    --verbose \
    --force_overwrite \
    --plugin MaveDB,file=/nfs/production/flicek/ensembl/variation/data/MaveDB/MaveDB_variants.tsv.gz

# count the number of MaveDB annotations
cd /hps/nobackup/flicek/ensembl/variation/fairbrot/MaveDB/test_with_plugin
grep -v '^#' mavedb_old_pipeline_out.vcf | grep 'urn:mavedb:' | wc -l
grep -v '^#' mavedb_old_pipeline_out.vcf | grep -o 'urn:mavedb:' | wc -l
```

This should give 39 lines where at least one URN ID is assigned, and 473 individual instances of URN IDs being assigned in the file. 

These numbers are a successful sense check for the new dataset, indicating that whilst the ballpark numbers are similar, a slight increase occurs in the new dataset due to the new data available in this version of MaveDB.

### Main issue that triggered extensive changes

For each experiment processed, the pipeline requires on 3 files: metadata.json, `mappings.json` and `scores.csv`. 
The `metadata.json` and `mappings.json` structures were different when downloaded on-the-fly via the API or directly from MaveDB. 

### Changes

- Added the `--from-files` argument (and optional data path args: `--mappings_path`, `--scores_path`, `--metadata_file`).
  - false allows the pipeline to be run in its original state, downloading files from the API, and true enables the new functionality and requiring that the user supplies pre-downloaded files. 

- Added `extract_metadata.nf` and `extract_metadata.py`. 
  - The API route pulls the metadata down individually for each URN ID. However, in the database dump, the metadata for all URN IDs is in a single file - `main.json`. 
  - extract_metadata extracts the current URN ID's metadata from main.json and formats it to match that of the API-yielded metadata structure.
  - I made this decision, as if the structure matches, less of the downstream code has to be changed, reducing the risk of introducing bugs.

- Added `import_from_files.nf` and `import_from_files.sh` to check for existence, load and wrangle the mappings and score files. 
  - The most important thing this step does is to check whether an incoming scores CSV file has invalid URN IDs that will prevent it from mapping downstream. 
  - Non-standard URN IDs arise from the variants being given temporary accessions (like `tmp:..-1`) before being made publicly available, and clearly propagated into the database dump of these files.   
  - If these are detected, they are replaced with the URN ID contained in the scores CSV filename. i.e `tmp123-09` > `urn:mavedb:00000000-a-1-09`.

- Added `map_scores_to_variants_fromfiles.py` and modified `mapping.nf` so that if `--from-files` is `true`, `map_scores_to_variants_fromfiles.py` will be utilised, but if `--from-files` is false, the original `map_scores_to_variants.py` will be used.
  - I decided to make a new version of the script/ new set of functions rather than modify the old one with if statements for cleanliness and to prevent the introduction of bugs.
  - Considerable changes had to be made to this file because unlike for the metadata, which I coerced into the API metadata shape with `extract_metadata.py`, the mappings.json file was too different in structure to the API format to make this viable.
  - This script additionally provides a lot of checks and emits warning messages for a host of scenarios/edge cases that arose within the new set of data available.

- Deprecated the `filter.nf` module in the `--from-files` `true` branch in favour of a direct channel filter.
  - The `filter.nf` subworkflow depended on output of the download step, which no longer occurs. So instead, the license is extracted from the metadata in `extract_metadata.py` and assessed by a channel filter.

- Slightly increased memory requirement in the `run_variant_recoder` module to prevent jobs running out of RAM.

- The `README` has been updated to reflect these changes.


nextflow/MaveDB/bin/extract_metadata.py
ainefairbrother, line 1: This is the original file as written by nuno. It's showing up as all new as my initial strategy was to adjust this file to allow either --from-files to be true or false and perform processing of the data accordingly. However, I then came to understand that the file structures were too divergent and so it made sense to have two separate .py scripts to handle the API route versus the data download route rather than a single chaotic script with confusing if/else statements.

nextflow/MaveDB/bin/map_scores_to_variants.py
ainefairbrother, line 1: This is a pylint auto change.

nakib103 (changes requested):
@ainefairbrother, thanks for the PR. I did an overview of the code changes and it overall looks good. Needs to do some testing by running the pipeline.

I pointed out some improvement areas. Please check them out.
Also from where did you download the mapping file? It was not in the [zenodo](https://zenodo.org/records/14172004) location.

ainefairbrother:
Thanks very much for your review @nakib103. Agree with and accepted your suggestions. 

RE the `mappings.json` file - yes, I also couldn't find where this was downloaded from originally. @jamie-m-a pointed me to the original [ticket](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-6559), wherein Sarah links to an [AWS bucket](https://mavedb-mapping.s3.us-east-2.amazonaws.com/20241023_mapped_score_sets.tar.gz) that contains it. You are correct, it is not available in Zenodo. I will add this to the docs.

nakib103 (changes requested):
Hi @ainefairbrother, 
Final bits of suggestions after running the pipeline.

nakib103 (approved):
Looks good, thanks @ainefairbrother.

Can you update the base branch to `main`.

PR #1159, opened 2025-03-25
Update clinvar script - set correct somatic status when creating variation feature

dglemos:


dglemos:
@jamie-m-a this PR is similar to https://github.com/Ensembl/ensembl-variation/pull/1154

PR #1162, opened 2025-04-28
Changes to the meta coord script for private db

olaaustine:
# Changes
Adding the option `--type` to argument. 
## Reason for this change 
It dies because private_db does not have variation in the name.

### DIFF
--- a/scripts/misc/update_meta_coord.pl
+++ b/scripts/misc/update_meta_coord.pl
@@ -51,7 +51,7 @@ my $opts = $cli_helper->process_args( $optsd, \&usage );
 my ($variation_dbas, $core_dbas);
 if (defined $opts->{dbname}) {
   $variation_dbas = get_dbas($opts);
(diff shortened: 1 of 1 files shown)


scripts/misc/update_meta_coord.pl
```diff
@@ -50,7 +51,7 @@
 my ($variation_dbas, $core_dbas);
 if (defined $opts->{dbname}) {
   $variation_dbas = get_dbas($opts);
-  die "Variation database required" unless ($opts->{dbname} =~ m/variation/);
+  die "Variation database required" unless ($opts->{dbname} =~ m/variation/ || $opts->{type} eq "privates");
```
dglemos, line 54: `$opts->{type} eq "privates"`

Don't you mean `private`?

olaaustine: Updated

dglemos (approved):

PR #1164, opened 2025-05-02
Update trim_sequences in utils - add parameter $strand

dglemos:
Ticket: [ENSVAR-6780](https://embl.atlassian.net/browse/ENSVAR-6780)

This PR adds a new option to trim_sequences()
New option: $strand
The strand is necessary when we input a HGVSc in VEP.

**Testing**
Run VEP with input `ENST00000357654:c.4729_4731delinsTAG`


nakib103:
@dglemos , in the output the HGVSc is shown as -
`ENST00000357654.9:c.4730_4731inv`
which says it has inverted the amino acids in position `4730` and `4731` which seems wrong.

PR #1168, opened 2025-06-27
Update ClinVar script to save ref allele

dglemos:
- Update the ClinVar script to save the ref allele as attrib_type `clinvar_ref_allele`.
- Fix bug in attrib variation_names. new variants have too many rsids, I updated the code to only include the supported length.

Ticket: [ENSVAR-5773](https://embl.atlassian.net/browse/ENSVAR-5773)


jamie-m-a (approved):
Seems to resolve the issue

PR #1170, opened 2025-07-21
Update vcf_config for sheep reference change

nakib103:
Sheep reference and consequently their name has changed. This should be reflected in vcf config.


jamie-m-a:
LGTM

PR #1173, opened 2025-07-30
schema patch

jamie-m-a:
Missed the schema patch...


nakib103:
merged to `main` and `release/116`

PR #1176, opened 2025-08-05
Add AlphaMissense and ESM1b score in protein function

nakib103:
https://embl.atlassian.net/browse/ENSVAR-6812

Pipeline update:
Update ProteinFunction pipeline to load AlphaMissense and ESM1b scores and prediction instead of MetaLR and MutationAssessor. This will be effective for dbNSFP version >= 5.0.

API update:
The API have added functionality to retrieve and store from these 2 predictors.

1. the prediction matrix now use 3 bits for prediction. Because Alphamissense from dbNSFP can have 5 different prediction hence cannot fit into 2 bits.
2. For ESM1b scores which can be between -24.538 to 6.937 the scores are converted from a (-50)-50 coord to 0-1 coord by adding 50 and dividing by 100 before storing. The reverse operation converts the score back to (-50)-50 coord. To keep accuracy in the decimal digit we round up the score to first decimal place.

Unit test update:
Because of the change to prediction bit occupying 2 bits to 3 bits, we needed to update the protein function matrix in test database - `modules/t/test-genome-DBs/homo_sapiens/variation/protein_function_predictions.txt`
The same has not been done for cache, partly due to complexity in generating those and because only one test is effected by this. The test data has been change for this test case so it can pass.
Besides above, test data has been added for dbNSFP 5.0 and 5.2 and the new predictors.

**Test**
pipeline - http://guihive.ebi.ac.uk:8080/versions/96/?driver=mysql&username=ensadmin&host=mysql-ens-var-prod-4&port=4694&dbname=snhossain_ehive_pf_dbsnfp_test_116_38_homo_sapiens&passwd=xxxxx
(the pipeline was not fully run because of codon unavailability)

The following database has been updated with the above pipeline -
@v4 `snhossain_homo_sapiens_variation_116_38`

Related PRs -
https://github.com/Ensembl/ensembl-webcode/pull/1100
https://github.com/Ensembl/public-plugins/pull/910
https://github.com/Ensembl/ensembl-glossary/pull/11

### DIFF
--- a/modules/Bio/EnsEMBL/Variation/DBSQL/vcf_config.json
+++ b/modules/Bio/EnsEMBL/Variation/DBSQL/vcf_config.json
@@ -1526,6 +1526,14 @@
             "type": "remote",
             "filename_template": "https://ftp.ensembl.org/pub/data_files/sus_scrofa/Sscrofa11.1/variation_annotation/ALL_pCADD-PHRED-scores.tsv.gz",
             "annotation_type": "cadd"
+        },
+        {
+            "id": "CADD_GRCg6a_whole_genome_SNVs",
+            "species": "gallus_gallus_gca000002315v5",
+            "assembly": "GRCg6a",
+            "type": "remote",
+            "filename_template": "https://ftp.ensembl.org/pub/data_files/gallus_gallus_gca000002315v5/GRCg6a/variation_annotation/chCADD_updated.tsv.gz",
+            "annotation_type": "cadd"
         }
     ]
 }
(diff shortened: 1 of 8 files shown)


modules/Bio/EnsEMBL/Variation/Utils/DbNSFPProteinFunctionAnnotation.pm
jamie-m-a, line 1: Version specific updates and new supporting methods look good.

modules/t/vep.t
jamie-m-a, line 1: Thanks for updating this test

jamie-m-a (changes requested):
Looks generally fine, some suggested changes - see individual comments

jamie-m-a (approved):
sprintf seems to have solved the problematic esm1b scores - thanks!

PR #1208, opened 2026-02-23
Fix gencc

ainefairbrother:
- Set external_id
- Remove auto G2P and Orphanet filtration


jamie-m-a:
used in prod already, mergeing to keep record accurate

PR #1212, opened 2026-03-25
Improve ld_vcf tool

EbiArnie:
- Enable -h help option
- Check for mandatory file for -f
- exit instead of abort

### DIFF
--- a/C_code/ld_vcf.c
+++ b/C_code/ld_vcf.c
@@ -527,7 +527,7 @@ int main(int argc, char *argv[]) {
     /* getopt_long stores the option index here. */
     int option_index = 0;
 
-    c = getopt_long (argc, argv, "f:l:r:s:w:v:n:p:", long_options, &option_index);
+    c = getopt_long (argc, argv, "f:l:r:s:w:v:n:p:h", long_options, &option_index);
 
     /* Detect the end of the options. */
     if (c == -1)
@@ -567,18 +567,26 @@ int main(int argc, char *argv[]) {
         variants_file = optarg;
         break;
 
+      case 'h':
+        fprintf(stderr, "Valid options are [flrswvnp]\n");
+        exit (1);
+
       case '?':
         /* getopt_long already printed an error message. */
         break;
 
       default:
-        abort ();
+        exit (1);
     }
   }
 
   if (numregions > MAX_REGIONS) {
     fprintf(stderr, "Number of maximum allowed regions exceeded: %d.\n", numregions);
     return EXIT_FAILURE;
+  }
+  if (filestr == NULL) {
+        fprintf(stderr, "<-f file> must be specified\n");
+        exit (1);
   }
 
   char *files[MAX_REGIONS];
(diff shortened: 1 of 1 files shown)


C_code/ld_vcf.c
```diff
@@ -567,19 +567,27 @@ int main(int argc, char *argv[]) {
         variants_file = optarg;
         break;
 
+      case 'h':
+        fprintf(stderr, "Valid options are [flrswvnp]\n");
```
ainefairbrother:
```suggestion
        usage(argv[0]);
```

C_code/ld_vcf.c
```diff
@@ -567,19 +567,27 @@ int main(int argc, char *argv[]) {
...
+        exit (1);
```
ainefairbrother:
```suggestion
        return EXIT_SUCCESS;
```

C_code/ld_vcf.c
ainefairbrother, line 541:
```suggestion
      {"help", no_argument, 0, 'h'},
      {0, 0, 0, 0}
```

ainefairbrother (changes requested):
This fixes the segfaults that were previously occurring. However,  `-h` should trigger the existing usage text which gives full example command, not new text. See below:

this PR `-h` output
```
Valid options are [flrswvnp]
```

suggested `-h` output
```
Usage: ld_vcf -f [input.vcf.gz] -r [chr:start-end] -l [optional_sample_list] (-g [input_two.vcf.gz] -s [chr:start-end]) > output.txt
```

I have added code suggestions to achieve the latter case. I also added a quality of life change - I updated the existing usage text to be more useful (L426).

C_code/ld_vcf.c
ainefairbrother:
```suggestion
  fprintf(stderr,
    "Usage: %s -f <input.vcf.gz> -r <chr:start-end> [options] > output.txt\n"
    "\n"
    "Required:\n"
    "  -f, --file <file>                Input VCF/BCF file\n"
    "  -r, --region <region>            Region to query, e.g. 1:1000-2000\n"
    "\n"
    "Optional:\n"
    "  -l, --samples <file>             File with samples to include\n"
    "  -n, --include_variants <file>    File listing variant IDs to include\n"
    "  -v, --variant <id>               Focal variant ID\n"
    "  -p, --var_position <int>         Focal variant position\n"
    "  -s, --numregions <int>           Number of file/region pairs to process\n"
    "  -w, --window <int>               Window size in base pairs\n"
    "  -h, --help                       Show this help and exit\n",
    prog
  );
```

ainefairbrother (changes requested):

C_code/ld_vcf.c
```diff
@@ -567,19 +567,27 @@ int main(int argc, char *argv[]) {
...
+
       case '?':
         /* getopt_long already printed an error message. */
         break;
 
       default:
-        abort ();
+        exit (1);
```
nakib103:
```suggestion
        return EXIT_FAILURE;
```

nakib103 (changes requested):

PR #1213, opened 2026-04-07
Update drosophila EVA source to match

nakib103:
The source for drosophila has not been updated to EVA release 8.


likhitha-surapaneni (approved):
LGTM

PR #1214, opened 2026-04-07
bugfix: HGVSg - the check length gets truncated

nakib103:
https://github.com/Ensembl/ensembl-vep/issues/1984

When shifting 3' for HGVSg we should skim through whole of downstream sequence (`post_seq`) instead of just the difference between the variation reference sequence (`seq_to_check`) and the downstream sequence. 

Generally the difference is only 50 bp added as extra to the total variant length in [_get_flank_seq](https://github.com/Ensembl/ensembl-variation/blob/23c76f60b1592e4df86159cf5530bdc326120c3d/modules/Bio/EnsEMBL/Variation/VariationFeature.pm#L1816) function.


likhitha-surapaneni (approved):
Tested and looks good to me. Thanks @nakib103

PR #1216, opened 2026-06-01
Fix/ld vcf2

nakib103:
replacement of [PR](https://github.com/Ensembl/ensembl-variation/pull/1212) against postreleasefix branch.


nakib103:
merged to `main` and `release/116`

PR #1219, opened 2026-08-17
Update pangenomes pipeline

ainefairbrother:
## Description

This PR brings the pangenomes pipeline up-to-date. A recent attempt to run it on the pangenomes assemblies revealed that some changes were required. I have detailed these below:

- Replace GTF support with GFF3 support
    - Filter for genes using ID=gene, and filter for transcripts from Parent=gene and transcript IDs
    - Handle GFF3 feature types
    - Rename feature labels to gene/transcript for compatibility with the previous output format and with VEP plugins
- Handle assembly accessions
    - Previously, the accession was grabbed from the input filename
    - The new inputs have generic names such as genes.gff3.gz, so the accession is supplied by the user
    - Use user-supplied accession to construct plugin output filenames
- Update download mode
    - Update the FTP download URL and handle new structure
    - Get the latest dated geneset for each assembly
    - Update the HGNC gene-symbol lookup URL
- Add manifest input mode
    - User can now supply a manifest TSV with the following cols: accession, fasta, gff3
    - Useful for local files stored in different locations, and for fasta/gff3 files with generic names
    - Do not use download mode when a manifest is supplied, but default to it when no manifest supplied
- Update the test so that it doesn't pick random regions - some of the haplotypes have more sparse annotations, and the test was picking regions with genuinely no annotation and failing

**Note:** 
The pipeline uses the DB to grab grch38 gene-GO and gene-phenotype associations, which are then projected to the haplotype assemblies using gene symbol matching. In future, the phenotype data store will serve the gene-phenotype look-ups, and an alternative source of gene-GO terms will be a TODO.

## Testing

Download fa and gff3 from: https://ftp.ebi.ac.uk/pub/ensemblorganisms/Homo_sapiens/GCA_018466835.1/

Create a manifest.tsv, like:
```tsv
accession     fasta   gff3
GCA_018466835.1       /path/to/unmasked.fa.gz        /path/to/genes.gff3.gz
```

Run with: 
```bash
nextflow run \
  ensembl-variation/nextflow/pangenomes \
  -profile slurm \
  --input_manifest manifest.tsv \
  --version 116 \
  --user xxxx \
  --host xxxx \
  --port xxxx \
  --outdir .
```

### DIFF
--- a/nextflow/pangenomes/bin/create_pangenomes_annotation.py
+++ b/nextflow/pangenomes/bin/create_pangenomes_annotation.py
@@ -7,7 +7,8 @@ import csv
 
 parser = argparse.ArgumentParser()
 parser.add_argument('--version', help="Release version", required=True)
-parser.add_argument('--gtf', help="Assembly-specific GFF/GTF", required=True)
+parser.add_argument('--accession', help="Assembly accession", required=True)
+parser.add_argument('--gff3', help="Assembly-specific GFF3", required=True)
 parser.add_argument('--outdir', default=".", help="Output directory")
 parser.add_argument('--gene_symbols', default=None,
                     help="Lookup table with two columns (gene symbols and Ensembl identifiers) ")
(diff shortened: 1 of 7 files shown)


jamie-m-a (changes requested):
HI @ainefairbrother 

Sorry for the late review. So I assume this has been run in prod by now? In that case probably fine to merge. Just a couple of minor comments:

Probably better to use softmasked.fasta to be consistent with web VEP.

Could we do something a bit smarter with memory use and retries (has it been profiled to see actual memory use through a run?), see nf-web-vep repo for (slightly) better memory use/ retry strategy

